PR #3, opened 2015-11-30
Fix trailing spaces in cluster name

### LINKED ISSUE #5: A cluster address with trailing spaces results in a host not found error
TRCLUSTER.DAT has names that have trailing spaces. When the program tries to connect to the clulster, the spaces present an error. This spaces should be trimmed from the end of the hostname before the connect attempt is made. I noticed that this is a particular issue if the hostname does NOT have :<port number> at the end.

ny4i:
Fixes #5 by replacing the extra spaces at the end with \hx0. There is code in uTelnet that reads the hostname from the rear looking for a :. I added code to look for a space and replace it with a null. Note if the beginning of the hostname has spaces, this could be an issue. Using the homegrown trim code did not work.


n4af:
Hi Tom -

I am not familiar with the pull request (but need to learn).

My top thought is we do not want to integrate this into 4.34.4 - I would
prefer you put it out as a new fix-- say open an issue in gm0gav tracker
and assign to 4.34.5 ?

I suppose I need to manually commit the changes to master and 4.34.4 to get
the crap outa logsubs1 also.

Good news is I have my pc fixed- bad news is my git local repo is still
hosed by crap in logsubs1.

73, Howie

http://n4af.net

PR #152, opened 2016-03-17
Issue149

ny4i:


ny4i:
Please check this out (with two radio mode too) and merge into MASTER if OK.

PR #156, opened 2016-03-19
Cw timing

ny4i:
Howie - This is just an experiment. Could you pull this code down and run it? I am curious if the ALT-D timing is any better with two K3s now. Thanks

# This can  now be merged if it works OK.
I added code to run better with my Icom so I can now test quite well and Alt-D works very well. I also removed some redundant SetSpeed calls that were generating too many CAT messages due to SpeedSync.

 I have some cleanup to do of the code. I just wanted to see if this will work before I make is a general implementation.

### DIFF
--- a/tr4w/src/trdos/LOGRADIO.PAS
+++ b/tr4w/src/trdos/LOGRADIO.PAS
@@ -37,7 +37,8 @@ uses
    Tree,
    VC,
    Windows,
-   utils_text;
+   utils_text,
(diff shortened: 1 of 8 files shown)


n4af:
I would suggest we merge--once i have put it into the rdxc, which starts
now.
it seems to work quite well, but needs more testing.

Great job ! I am going to try cwbc in rudxc

73, Howie

http://n4af.net

Issue #218, opened 2016-12-28
Add support for FT991, 991A and 891


ny4i:
Some commands (FA for example) have changed. Identify the changes and code up a new type of Yaesu with the longer FA field and whatever other changes are found.

ny4i:
This was a good segue to get me back to coding. Plus I introduced a bug when I put 991 in the list but it would not work. This is actually easier to do that I thought as I'm almost done. We will of course need to have Yaesu rig owners test it. Preferably someone with an 891 and an older Yaesu to make sure I did not break anything.

n4af:
Thanks Tom !!!!

ny4i:
User reports that this now works on the FT891

ny4i:
I forgot to do a Global search for a radio to make sure I added FT891 and FT991 to every place where a Yaesu rig was specified. I found two more commands that had something specific for the radio (UP and DN command). This fixes that.

PR #220, opened 2016-12-29
Fixes Issue #218 to add FT891 and fix FT991

ny4i:
Howie - Can you compile this into a beta and ask Sasa to test it. I do not have either of these radios. As I'm not completely setup here yet, you might want to make sure this does not break your regular radio stuff. The code was very specific to Yaesu so I doubt that would have an impact. If we know of another Yaesu user with FT5000 or the like, that would be a good test too. — Thanks

### DIFF
--- a/tr4w/src/VC.pas
+++ b/tr4w/src/VC.pas
@@ -946,7 +946,8 @@ type
     FT840,
     FT847,
     FT857, {UA3DPX}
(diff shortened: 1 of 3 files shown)


n4af:
Thanks Tom -

http://tr4w.net/beta/tr4w_setup_4_55.Beta.exe

PR #238, opened 2017-02-14
This change fixes #222 to add Winter Field Day as a valid contest type.

### LINKED ISSUE #222: Add support for Winter Field Day
http://www.winterfieldday.com/rules/

This is almost identical to Field Day but they categories are different (O, H, I).

ny4i:
Howie - Please take a look and just make sure I did not break anything. I would test picking a contest from the top of the list and the bottom and making sure the questions (such as your state, etc) line up. That is how I tested it. This was to add support for Winter Field Day.

### DIFF
--- a/tr4w/src/VC.pas
+++ b/tr4w/src/VC.pas
@@ -915,7 +915,8 @@ type
     ARRL_RTTY_ROUNDUP,
     CQIR,
     WWIH,
-    ALRS_UA1DZ_CUP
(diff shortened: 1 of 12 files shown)


n4af:
Hi Tom -

I have tested 4.56.9.
Best i can tell it looks good.
You might want to post to tr4w Support forum to get further testing ?
http://tr4w.net/4.56/tr4w_setup_4_56.9.exe

Thanks, Howie

PR #411, opened 2020-06-30
Fixed error with FD contest name

ny4i:
a prior change set the Field Day contest name to ARRL-FIELD-DAY. This changed the internal name from ARRL-FD. That caused an error with prior CFG files. I had previously added code to use an ADIFName and a CABName for display purposes. This change also updated some ADIF names but more still need to be done. I also had an Indy issue with a recently compiled version of Indy for tHttp. That was not really being used so I removed it from the code for now. Since I was int he area with the tHttp issue, I also updated the code that open the rules on the WA7BNM website.

Note I also moved .map files to my .gitignore file. We might want to do that in repo too unless they are needed. I thought they were generated upon compile.


ok1rr:
It seems that in 4.89.7 appears decimal comma instead decimal point in 
the FREQ field in ADIF export.
<FREQ:6>7,0451 ... 4.89.7
<FREQ:8>7.029050 ... 4.88.3

Any changes there?

73,
Martin, OK1RR

ny4i:
I will take a look. Yes there were significant changes in the area of ADIF Export but I did not change the generation of the frequency. It appears it is using what I assume is your localization standard of a comma.

I will check.

Tom NY4I
Principal Solutions Architect
727-437-2771

ny4i:
I found it. I have been switching to the standard Delphi Format statement and that used the localized decimal separator. I overrode that in this area and am going to switch my localization to test it.

Tom
Principal Solutions Architect
727-437-2771

ny4i:
All fixed and tested withy system set to use a common as the separator and verified it appears as a decimal character (.).

Thanks for the report Martin and sorry I forgot to test this part. I will make sure I change the localization when I test. Best thing is for me to make a separate Windows VM to test set for Spain (so I can read the prompts). 73...

Howie - I just added it to PostUtils in Master.

Tom
Principal Solutions Architect
727-437-2771

n4af:
Thanks Tom -

I did another pull which updated postunit:
--->
From github.com:n4af/TR4W
 * branch            master     -> FETCH_HEAD
   cf32908..d871857  master     -> origin/master
Updating cf32908..d871857
Fast-forward
 tr4w/src/trdos/PostUnit.PAS | 2 ++
 1 file changed, 2 insertions(+)

Going to generate 4.90 this morning.

73, Howie

PR #412, opened 2020-07-02
Telnet connect not in list

### LINKED ISSUE #67: Need better error handling when cannot connect to cluster
If we cannot connect to the cluster for some reason, the error messages are not very good. Sometimes we get a Windows WinSock error or just Command Completed Successfully. This should be beef'ed up and return better, more descriptive errors. It might not hurt to check the server or IP via a ping to test connectivity to the cluster

### LINKED ISSUE #392: TR4W DX Cluster not automatically connecting to the server if it is not listed in trcluster.dat
Following statement in .cfg-file does not open the connection automatically to the telnet server. In order to make it work you need to add the telnet server first to trcluster.dat-file.

;DX-cluster
;----------
TELNET SERVER = ham.connect.fi:7300
CONNECTION AT STARTUP = TRUE
CONNECTION COMMAND = OH6QU

ny4i:
The fixes #392 .  Now an item in the cluster config is added to the server drop-down list at run time but does not modify TRCLUSTER file. Also added better messages if the the cluster cannot connect so this also fixes #67 .


ny4i:
Reopen this as it appears something was missed. It might have been when I Merged master into the branch (above). That is what happenes when I try to do multiple patches at one time.

PR #468, opened 2020-12-09
Update uRadioPolling.pas

Dan-K0TI:
Revert delay in polling code to pre 4.73.5 number to fix issue with K3/MicroHam MK2R. This is a tested fix at the W0ZT station. 
I'm not sure it needs to go as high as 80, something shorter might work, as I'm don't have easy access to the station (with covid and all) I am unable to easily test other values.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,2 +1,3 @@
 # TR4W
 TRLOG 4 Windows free amateur radio logging application
+The Fork is just for debugging of K3 polling issues.
--- a/tr4w/build/make_setup_file.bat
+++ b/tr4w/build/make_setup_file.bat
@@ -1,4 +1,4 @@
 rem UpResource.exe
 upx.exe ..\target\tr4w.exe --lzma
-"D:\Program Files\NSIS\makensisw.exe" full.nsi
+"C:\Program Files (x86)\NSIS\makensisw.exe" full.nsi
 pause
--- a/tr4w/src/uRadioPolling.pas
+++ b/tr4w/src/uRadioPolling.pas
@@ -163,7 +163,7 @@ begin
             BufferNotChanged := 0;
 
             NextWait:
-            Sleep(80); // 4.73.5 was set to 40  Revert to 80 to fix K3/Microham (K0TI)
+            Sleep(80);  // 4.73.5 was set to 40  Revert to 80 to fix K3/Microham (K0TI)
             ClearCommError(rig^.tCATPortHandle, Errs, @stat);
             if stat.cbInQue > BytesInBuffer then
                begin
--- a/tr4w/src/uRadioPolling.pas
+++ b/tr4w/src/uRadioPolling.pas
@@ -163,7 +163,7 @@ begin
             BufferNotChanged := 0;
 
             NextWait:
-            Sleep(10); // 4.73.5 was set to 40
+            Sleep(80); // 4.73.5 was set to 40  Revert to 80 to fix K3/Microham (K0TI)
             ClearCommError(rig^.tCATPortHandle, Errs, @stat);
             if stat.cbInQue > BytesInBuffer then
                begin


Dan-K0TI:
I had not intended that the readme.md change I made should become part of the pull request. Only the change to 
uRadioPolling.pas. 
Should I close this request and start another one ?.

n4af:
integrated: 4.95.3

Dan-K0TI:
Howie, thanks for that. I'll have Steve test that when it get's released.
That merge changed your Readme.md file however, not my intention, you might want to change that back. Sorry.
I have to learn more about how all this interacts on GitHub.

n4af:
no problems Dan.
Hope it fixes things.

73, Howie

ny4i:
Dan,

What issue did you see that required increasing the polling delay? I ask as I did the FT Roundup this weekend and did not see any issues. Granted, I was not changing frequency much and the only thing that really changes is the XMIT ON status.

On a separate note…great to have another committer on the project.

Tom NY4I

Principal Solutions Architect
727-437-2771

Dan-K0TI:
Tom,
The issue was brought to my attention by Steve K0SR, at the W0ZT contest station. He had been struggling with it for some time.
This is with K3’s (SO2R) and the microham MK2R+ controller. The frequency display’s in tr4w for both radios keeps blinking on and off and the microkeyer gets messed up also so it doesn’t know where the radios are. And there are all kinds of errors in the serial port log in the microkeyer. I also have a MK2R here but only have a KX3 to test but I’m able to see a similar issue.
Just changing the FreqPollTime doesn't fix it.

I saw a note in changes made in 4.73.5 about speeding up the K3 polling and that seemed to be the clue as to what might have changed.

I built a test version that for Steve yesterday based on 4.94.2 and that fixed it. 
However now Steve just reported that the 4.95.3 build that I just had him try doesn’t fix it. But I have to dig into that a bit more.
This my take a bit to sort out and I’m going to be gone next week so this might go dormant for a bit until I can get back on it.

Thanks for the welcome. 

Dan

Dan-K0TI:
OK False alarm on the fix not working. He didn't get the right version at first. The 4.95.3 build does fix this.
When I get time in 10 days or so I'm going to investigate if the fact that it sets the interface to 2 stop bits is a factor in this.
The elecraft radios are supposed to be 1 stop bit. 

Thanks Howie and Tom

n4af:
if u have not already tried, ensure
NO POLL DURING PTT is set to true...

n4af:
ok. i run so2r/K3's so if the poll wait does not need to lengthen for
MicroH, would like to change it back.

73, Howie

Dan-K0TI:
Howie,
 I do think that it doesn't need to be 80 , but I can't really test for that right now. I've built myself a version that will allow me to vary that wait time on the fly and I should be at the site for Stew Perry so I'll test that then and see what the lowest is that it can go with that setup. There's K4's on order for there so that might change things :)

Dan

n4af:
FB. I suspect the K4 runs the same firmware for polling.
Also, the polling speed may have some effect.

73, Howie

ny4i:
OK. A while ago I added some code to not refresh the display if the frequency did not change. This stopped the flashing of the frequency when it did not  change. I have a MK2R+ as well and its frequency is coming directly from the radio before the data is sent to the program so I am not sure why the program would have effect on that.

BTW, there is a log4j type logging mechanism in the program. You set the CTRL-J DEBUG option to the level you want and in the program, use logger.debug(‘Hello’) or logger.trace(’This is a message %s’,[sData]);

The debug log is in the program directory. That is helpful while tracking down issues like this one.

Tom NY4I
Principal Solutions Architect
727-437-2771

Dan-K0TI:
Tom,
 Perhaps saying the frequency was flashing is misleading. Better said the frequency was going away, even the one displayed in the main form above the radio name. The VFO A and or B would go blank, maybe every second or so, dependent on the freqpollrate as I recall. Plus if you watched the MK2R log scroll by it would hesitate every so often like it had missed a message. With the fix it's very smooth, I can see that difference here with the KX3. I had captured the time between TX/RX messages and they would jump all over the place. The MK2R would loose track of one radio or the other and go into lockout mode (busy light on). 
Attached are MK2R log files for 4.94.X which was broke and 4.72.3 which was that last one I could find that worked.
As you can see even the MK2R was getting errors. Turning off the MK2R polls did not help nor did changing freqpollrate.
I haven't studied the code in that section quit enough yet to understand fully what that's going on there but changing that
waittime makes a difference.

I understand wanting to update as fast as possible, but in this case it overloads something. Also the MK2R needs to poll for  information every so often so some consideration has to be given for that. The K3 reference manual says that the latency could be up to 100ms, with a band change up to 500ms. So going much faster than that might be pushing things.

As I stated in an earlier post I want to investigate if having it set to 2 stop bits might have an effect. It would not affect the data going to the radio, but if the data coming from the radio is 1 stop bit and the computer is expecting 2 that should cause problems. I'm actually surprised it works as well as it does. Also if you notice the MK2R changes to 2 stop bits when the computer opens the port at 2 stop bits, so the data it gets could be affected by this. There is more to learn about this yet.

[SR4.94.log](https://github.com/n4af/TR4W/files/5668854/SR4.94.log)
[SR4.72.log](https://github.com/n4af/TR4W/files/5668855/SR4.72.log)

Thanks for the debug info. I saw some references to it, didn't investigate how to make it play.

Dan K0TI

ny4i:
For what it is worth, I do plan on adding code to talk the the K4 via Ethernet/TCP ( when mine arrives that is). 

Tom

Principal Solutions Architect
727-437-2771

ny4i:
All, would it be helpful to make the poll delay a parameter we can set with a CTRL-J setting?

Tom NY4I  

Principal Solutions Architect
727-437-2771

PR #469, opened 2020-12-11
Polling update to Kenwood/K3

Dan-K0TI:
This changes the K3 polling code so that a long wait time is not required. This code checks for new data every 10 ms and if it gets the number of bytes expected it moves on without any further waiting, otherwise it will wait for as long as 1000ms for response from the radio. This should give the best of both worlds. This has been tested extensively on the KX3 here at baud rates from 4800 up to 38400. It has not been tested on K3, so that needs to be tested. 
I found on the KX3 that the normal time for a radio response to be 30-40 ms, but if your tuning the radio it's about 100-150, and a band change can take 600 or more.

The same code structure is in the Orion polling section but as I don't have a radio to test I didn't update that. 

This also fixes that Readme.md that I broke before. My son was teaching me git workflows so he generated that change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,2 +1,3 @@
 # TR4W
 TRLOG 4 Windows free amateur radio logging application
+The Fork is just for debugging of K3 polling issues.
--- a/tr4w/src/uRadioPolling.pas
+++ b/tr4w/src/uRadioPolling.pas
@@ -202,24 +202,6 @@ begin
                 logger.warn('Buffer Bytes -' + inttostr(BytesInBuffer));
               end;
 
-{
-            NextWait:
-            Sleep(80);  // 4.73.5 was set to 40  Revert to 80 to fix K3/Microham (K0TI)
-            ClearCommError(rig^.tCATPortHandle, Errs, @stat);
-            if stat.cbInQue > BytesInBuffer then
-               begin
-                  BytesInBuffer := stat.cbInQue;
-                  goto NextWait;
-               end
-            else
-               begin
-                  inc(BufferNotChanged);
-                  if BufferNotChanged < 3 then
-                     goto NextWait;
-               end;
-}
-
-
             if BytesInBuffer > 0 then
                begin
                   inc(NumberOfSucceffulPolls);
--- a/tr4w/src/uRadioPolling.pas
+++ b/tr4w/src/uRadioPolling.pas
@@ -121,6 +121,9 @@ var
    Errs: DWORD;
    BytesInBuffer: integer;
    BufferNotChanged: integer;
+   RadioWaitTime: integer;
+   RadioTimeoutTime: integer;
+   RadioWaitLoops: integer;
    NumberOfSucceffulPolls: integer;
    i: integer;
    TempCommand: tKenwoodCommands;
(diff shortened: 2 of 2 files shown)


n4af:
testing this on dual k3 so2r.
The plus is the response of TR4W to tuning S&P rig is excellent !
One negative is the active/run rig, whilst sending, will flicker to indicate freq of S&P rig. But it returns to the proper run freq display.
I will need to put this into the weekly CWT to beat on it, but the response to change in tune freq on S&P rig seems exceptional.

Dan-K0TI:
I wonder why the freq of radio 2 would get displayed the radio 1 display, is that what your saying ?.
Not sure I understand that at all.

n4af:
For me, just tuning S&P does not affect Run. 
But when I CQ AND tune S&P the displayed S&P will (momentarily) display run freq.
RS232 set on both radios to 19200.

Dan-K0TI:
Which frequency display are we talking about the one in the main window or the one in the Radio 1 / Radio 2 windows

n4af:
rr. Not sure why it happens, but I asked N5AW if he could try it in his
2Xk3 so2r setup.

NO POLL DURING PTT makes no difference.

n4af:
Main window...

Dan-K0TI:
Ok, I haven 't figured out yet in the code where those get updated at.

ny4i:
All the main window fields are prefaced with mwe_. If I recall, the ActiveRadioPtr (which is pointing to the Radio1 object). Has the VFO record updated and then in the refresh routines, them we field is updated with the value from the Radio object.

Tom NY4I
Principal Solutions Architect
727-437-2771

Dan-K0TI:
I've look over my code changes again and do see where it could do this. Only thing I'm thinking is that it's probably going thru there faster now and we have some race condition someplace. Maybe try it with a higher frequencypoolrate and see if it settles down. Still thinking about it.

n4af:
I put the faster polling thru both CWT and NS this week. All seemed FB running SO2R*K3s's.
No time in heat of contest to look for the flickering but I think if it stayed on band of second radio, I would notice.

Dan-K0TI:
That's good to hear. I've still been thinking about what would cause that, have tried to reproduce it with my limited setup here with no luck, but I have some ideas I want to look into next week at home. I see you set it up to allow the freqpollrate to adjust that speed, have you tested to see if that make's any difference  ?. I also found another issue in the ProcessFilteredStatus procedure with the use of an uninitialized variable, however I don't know if it would ever be an real issue, but I'll generate a pull request for it next week.

n4af:
Hi Dan -

Yeah, I did try exploiting Freq Poll Rate, but it d/n make a lot of
difference.

73, Howie

Dan-K0TI:
Yes I figured it was deeper than that. I suspect that the radio object (either the whole object or just the frequency value) for the  radio1 is getting clobbered while polling radio2 and seeing as radio1 isn't being polled while transmitting you see it on the screen. As soon as they both are updating again it goes back to normal. If I could replicate it I could track it down. Maybe Tom has some ideas on that.

ny4i:
Just off the top of my head, there  is the idea of ActiveRadioPtr which points to the active. I wonder if there is a case where the pointer (which might point to Radio2) is being used instead of Radio1 in the poll cycle.

Again, just with 5 seconds of thought.

Tom
Principal Solutions Architect
727-437-2771

Dan-K0TI:
That's along the lines I was thinking, seeing as it's multi threaded (right ?, I haven't dug into ALL the code yet)  perhaps something isn't thread safe or its just getting clobbered due to timing. That's the kind of things I was going to try and investigate next week

PR #470, opened 2020-12-20
Update uRadioPolling.pas

Dan-K0TI:
Move location of variable dif assignment so it's used for both active and inactive radios The inactiveradio portion of this if-then-else section was using an unasssigned value for  dif

K0TI 12/19/2020

### DIFF
--- a/tr4w/src/uRadioPolling.pas
+++ b/tr4w/src/uRadioPolling.pas
@@ -2554,9 +2554,10 @@ begin
                      end;
                end;
          end;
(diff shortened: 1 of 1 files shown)


n4af:
No change in the flicker.
Since I am only one experiencing this, it may be prudent to wait for more feedback.

Dan-K0TI:
I didn't expect this to change anything. For almost every case thru the procedure 'dif' would have been > 0 anyway so the result of the test didn't change. At least the code is cleaned up so there is no unexpected behavior here.

PR #472, opened 2021-02-01
Fix to Winter Field Day scoring and Cab generation

ny4i:
I use the power version to calculate the score so the user only needs to add in their bonus points. I also updated the QSOPoint calculation to include Phone and FM as modes that are one point. Previously it was just phone so FM ended up with 2 points. Field Day had the same bug.


ny4i:
Howie -- if you could generate this and test with a few other contests to make sure Cab file looks ok. I also need an installer to have a few friends that used TR4W for Winter Field Day use to generate their Cab files. Thanks.

n4af:
https://tr4w.net/4.97/tr4w_setup_4_97.1.exe

PR #486, opened 2021-02-23
Added additional debug logic to BandMap

ny4i:
Trying to trace an issue with a user clicking on the bandmap but the frequency not changing.


ny4i:
Have one more change for this so cancelling this PULL request

PR #487, opened 2021-02-23
Band map debug

ny4i:
Added more BandMap debug messages as well as a trace to see the data coming back from the radio. This way when the user clicks on a spot, we can see if the radio is responding properly to the command. This will show for example, if the radio command is not implemented as is the case if an IC7600 or IC7700 does not have the Version 2 and 2.1 firmware respectively.

### DIFF
--- a/tr4w/src/uBandmap.pas
+++ b/tr4w/src/uBandmap.pas
@@ -393,13 +393,19 @@ begin
           LBN_DBLCLK:
             begin
               TempInt := GetBMSelItemData;
-              if TempInt = LB_ERR then Exit;
+              if TempInt = LB_ERR then
+                 begin
+                 logger.Trace('In BandMap::BandmapDlgProc, GetBMSelItemData = LB_ERR so exiting without changing radio');
+                 Exit;
+                 end;
               if  QSYInactiveRadio and  TwoRadioMode then          // 4.92.1                         //Gav 4.37.12
                 begin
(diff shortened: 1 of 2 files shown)


ny4i:
Valery will need the link to the this build, ask the chaps with the 7600 and 7700 to try this version and ensure DEBUG level is set to TRACE.

Afterwards, we need the trace log back for analysis.

Thanks,

Tom NY4I

Principal Solutions Architect
727-437-2771

n4af:
i wonder if the compile error against **utils_text** is caused by a bad call from the added code ?
https://github.com/n4af/TR4W/commit/a90167f19e5b9c4d79ac5d88c68f176a1025fc22

ny4i:
That seems logical. No changes to UTILS_TEXT that I recall though. Did you pull all the code from MASTER and the PULL REQUEST to compile it versus relying what was on your system?

Pulling MASTER without the pull request and trying to compile it would be a good test.

Tom
Principal Solutions Architect
727-437-2771

n4af:
yeah, did the pull and verified the code.
always possible something was already clobbered. I will restore back to
last good compile...

n4af:
it appears something is setup wrong with my compiler. i get the same issue restoring to the last and next to last releases.

n4af:
Finally located the problem as an old copy of tr4w.dpr in the **wrong director**y.
Interesting as I tried restoring the ENTIRE source directory with same problem, so the current project pointer must be in other than the dirve I maintains source code in.

PR #491, opened 2021-03-02
Fixed mode setting

ny4i:
Fixed mode setting for newer Icoms (Command 26 support) and Kenwood/K3.


n4af:
Thanks Tom.
Fixed 4.97.11

PR #522, opened 2021-07-01
Refactored the call window commands and added a few

ny4i:
I added OPON to change operators (same command as N1MM for ops that use both) and also added CWON and CWOFF commands. I also changed the code to a case statement for readability.

I also made it so when ALT-K is used the CW speed changes to NO CW so the user has an indication that no CW will be sent. This is same display in Toggle CW. Toggle CW checked just CWEnabled instead of CWEnable.
We seem to have two Booleans that indicate if CW is auto sent. This should be changed to only use a single variable. Otherwise it is easy to confuse.

### DIFF
--- a/tr4w/src/MainUnit.pas
+++ b/tr4w/src/MainUnit.pas
@@ -5897,45 +5897,77 @@ end;
 
 function CheckCommandInCallsignWindow: boolean;
 begin
-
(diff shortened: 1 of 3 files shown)


n4af:
Pushed to github as 4.101.7

PR #523, opened 2021-07-03
Added UDP command

### LINKED ISSUE #382: Add option to rebroadcast all QSOs to UDP network
In case whatever reading the QSOs via UDP gets lost, add an option to rebroadcast all the QSOs. This has to also send the delete UDP message followed by the Add contact message. This should be done in a thread in case of a large log.

ny4i:
New command in callsign window that will resend the UDP contacts. Note it does not send a ContactDelete message first though.

Enter UDP to resend contacts. This is a rare command so I did it this way rather than a menu.

Also, noticed a counter needed to be initialized in ImportFromADIF.

This closes #382 as well


n4af:
uploaded as 4.102.1

PR #525, opened 2021-07-06
Fix wsjtx field day

### LINKED ISSUE #524: WSJT contact is not entered due to missing domestic QTH in Field Day
It appears something broke in the WSJT-X interface for Field Day. When the contact is logged, it is not picked up as an error about the Domestic QTH missing is displayed.

ny4i:
Quite a few fixes for WSJT-X since I broke it for Field Day. Fixes #524 plus others.

### DIFF
--- a/tr4w/src/MainUnit.pas
+++ b/tr4w/src/MainUnit.pas
@@ -6438,19 +6438,6 @@ begin
                   tAdifGRIDSQUARE:
                      begin
                      gridSquare := fieldValue;
-                     if Length(exch.QTHString) = 0 then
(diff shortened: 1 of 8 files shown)


ny4i:
This requires testing of a few ADIF imports to make sure they are OK. I should create a few ADIF imports that I can run and check the import and export matches what is expected as an automated test case.

n4af:
This should be integrated into 4.102.2
<https://tr4w.net/4.102/tr4w_setup_4_102.2.exe>

Tks, Howie

PR #543, opened 2021-11-12
Update LOGRADIO.PAS

ny4i:
The code to trace output was not called if the length was 3, 6 or 7. This may have excluded needed strings for debug.
Note we poll far to often. Consider reducing poll frequency.

Also, only tracing for Icom radios is in hex. Most other radios are in text format (some older radios may not be).


n4af:
Well, I think we ended up with a greatly improved Debug Trace that will prove useful in the future.
Thanks, howie

PR #582, opened 2022-02-21
Added a manifest resource so version shows Windows 10 and 11 properly.

ny4i:
Added a manifest file so the Windows version is returned properly and a class to determine the OS to finer granularity.

This can be merged into MASTER. This will be my final patch this month.


n4af:
[4.108.10](https://tr4w.net/4.108/tr4w_setup_4_108.10.exe)

PR #587, opened 2022-03-10
The code in SendCW was not processed when sending to a network

### LINKED ISSUE #585: On a K4, when a preprogrammed message contains > (RIT clear), program crashes

ny4i:
A common theme was to short-circuit the specific function routine like SendCW to just send the message to the Networked radio. SendCW had important code that needed to be run even in the network case. So I added network code to AddToOutputBuffer. This fixed the CW. Closes #585

Ready to be built so Martin can test. Thanks

### DIFF
--- a/tr4w/src/trdos/LOGRADIO.PAS
+++ b/tr4w/src/trdos/LOGRADIO.PAS
@@ -1374,8 +1374,16 @@ begin
       nNumberOfBytesToWrite);
    Inc(CommandsBufferPointer, nNumberOfBytesToWrite);
 {$ELSE}
-  WriteToCATPort(Buffer^, nNumberOfBytesToWrite);
-  DebugMsg(Buffer);
+  if Self.tNetObject <> nil then
+     begin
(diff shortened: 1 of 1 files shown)


n4af:
Issue 585 is [integrated into 4.109.2](https://tr4w.net/4.109/tr4w_setup_4_109.2.exe)

ok1rr:
The 4.109.2 works but the AutoSend option has some quirks. When typing 
in the call, the sending option starts OK however tha spaces in the rest 
(while sending) has the problem with unnaturally long spaces like before 
(the rest of the call is torn to particular characters). The cursor does 
not jump into exchange field when the call sending is finished. So, 
AutoSend does NOT work.

With AutoSend off the program works as expected.

73,
Martin, OK1RR

Dne 10. 03. 22 v 12:56 Howie Hoyt napsal(a):

n4af:
Assign issue 619 <https://github.com/n4af/TR4W/issues/619>

n4af:
Hello Martin -

Does this occur under Windows ?
Do you have *AUTO CALL TERMINATE* set to TRUE ???

73,  Howie

PR #598, opened 2022-05-14
The update was actually in the ADIF parser in MainUnit

ny4i:
Added the code to the case statement to move the gridsquare as required.

Please build this as it is required for ARRL-DIGI.


n4af:
fix 4.111.1

PR #679, opened 2023-07-26
Issue 678 Yaesu serial CW not working

ny4i:
This has been tested on a Yaesu FT991A with a serial port via USB. This should be tested by the two people that had an issue with the CW on newer Yaesu radios. 

This should also be tested by everyone else that uses serial port CW (not WinKeyer or CWByCAT).


n4af:
Hi Cliff -

Pls try NY4I Yaesu fix <https://www.tr4w.net/4.124/tr4w_setup_4_124.12.exe>and
let us know if that fixed it ?

Thanks Howie

Issue #446, opened 2020-08-28
Add support for Expert Electronics Universal Transceiver Interface


ny4i:
https://github.com/maksimus1210/TCI/blob/master/doc/TCI_interface_1.4.pdf

Requested by Nikolai R3KM

This is a network interface so a socket has to be created with Indy.

ny4i:
Will add to be supported only by HamLib as it has an interface to TCI

ny4i:
Once the hamlib build is done, this should be able to use the HamLib interface. Just select the radio (at the end of the drop-down) and check the box to use HamLib.

ny4i:
Add a native implementation tothe radio factory. Her eis more reference info: https://pure-editions.com/on7off/TCI-Remote/TCI_Protocol_Reference.html

Note we are onlyu interested in the current functions TR4W supports. Audio processing, IQ data is outside of our scope.

PR #684, opened 2023-08-22
HamLib Implementation

### LINKED ISSUE #676: Implement support for HamLIb
Add support to allow radio control to be done by hamlib. This means an exception on just about all the radio code to handle the hamlib call and move on rather than send actual over the serial or network line.

ny4i:
This is a significant upgrade to TR4W to add hamlib support. Hamlib is the ham radio control library that WSJT-X and several other programs use. This allows us to automatically support radios that hamlib already supports such as the Expert Interface TCI requested in issue #446 . This also adds support for the Yaesu 757GXII and the Ten Tec Omni. The Flex should also work over the network.

There is also support to add a radio that is not specifically in the list but the hamlib ID is known as hamlib adds it. This allows hamlib to stay ahead of TR4W. At the moment, setting the hamlibID would have to be set in the INI file if TR4W's radio setup windows does not mention the radio.

Note this only works at the time for a single radio. I have yet to add the code to run two instances of rigctld but that will be easy once I get some feedback.

MY main concern is to make sure I have not broken anything. In all the LogRadio RadioObject functions, I just added code at the top to call HamLib if set. If not hamlib, then the regular rig control is set to run. This is similar to how I did the K4 network code.

CW by CAT should work for any radios that support it as does PlayMessage. Split should work as well. I tested tuning from the bandmap among other things so this should be ready to go but I do expect bug reports as people test it. Mainly the issue may be if a feature is used that the radio does not support through HamLib (direct VFOB control is one example that some rigs do not support--looking at you Yaesu).

Once built, I can send an announcement to the mailing list explaining how to use some of the features along with the link to the code.

This closes #676.


n4af:
Great job Tom !

https://tr4w.net/4.125/tr4w_setup_4_125.1.exe

ny4i:
Howie - I assume you are waiting for testing before you merge this into MASTER? Good plan if so...

n4af:
Hi Tom -

No, the master should include all the changes ??

ny4i:
Ok. Well the pull request was still active so it did not appear merged. TomPrincipal Solutions ArchitectBetter Software Solutions, Inc. 727-437-2771On Aug 23, 2023, at 6:30 AM, Howie Hoyt ***@***.***> wrote:﻿
Hi Tom -



No, the master should include all the changes ??



On Tue, Aug 22, 2023 at 5:21 PM Tom Schaefer ***@***.***>

wrote:

PR #701, opened 2023-10-15
Fixed up UDP band issue

ny4i:
Fixed an issue where the band being sent in the UDP Contact record was shifted due to an array ordering issue.

Also fixed an issue where and if the band had a decimal (1.8, 3.5, 5.4), it was not localized as per the N1MM UDP documentation.

Also fixed a related issue where the DecimnalSeparator was set to . incorrectly when sending the whole log to UDP.

Added Try..Finally code around the kludge to reset the DecimalSeparator to ensure it is set back to the proper value for the locale.

### DIFF
--- a/tr4w/src/MainUnit.pas
+++ b/tr4w/src/MainUnit.pas
@@ -6867,6 +6867,7 @@ var
   tempState: string;
   tempVE_Prov: string;
   tempARRL_Sect: string;
+  saveDecimalSeparator: char;
 
 begin
   lookingForFieldName := false;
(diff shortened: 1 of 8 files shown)


ny4i:
Note here is a simple UDP script to display received UDP messages:

```python
import socket

#UDP_IP_REC = "127.0.0.1"
UDP_PORT_REC = 12060

sock_rec = socket.socket(socket.AF_INET, # Internet
                        socket.SOCK_DGRAM) # UDP

sock_rec.bind(('', UDP_PORT_REC))

while True:
    data, addr = sock_rec.recvfrom(4096) # buffer size is 1024 bytes
    print("received message: %s" % data)
```

Note also that the DEBUG log setting will log the UDP output to the TR4W log file

ny4i:
Howie - I believe you did pull this request into the MASTER but didn't resolve the pull request.  This was the fix for the UDP band being wrong. The op that reported it said it was fixed so as long as this was merged into MASTER, this can be closed.

PR #720, opened 2024-01-29
Copy and paste error now fixed in UDP Port parameters

### LINKED ISSUE #715: UDP Contact Port setting is overridden by the default.
TR4W.INI has the line: UDP BROADCAST PORT CONTACT=12061 but when the program runs, it still shows in the CFG window as the default 120650./ Note this happened for one of three computers. The version was 4.130.2

Check that the default is not being set after the line is read from the config file. Test the other options as well.

ny4i:
BroadcastPortLookup was overwriting ContactPort. This fixes #715


n4af:
[fix 4.130.3](https://www.tr4w.net/4.130/tr4w_setup_4_130.3.exe)

PR #723, opened 2024-02-05
Fix wfd import from n1mm

ny4i:
Added importing class and added code to put the exchnage1 ADIF field from N1MM into the proper FOC Num field in the contest exchange.


n4af:
fix [4.131.2](https://www.tr4w.net/4.131/tr4w_setup_4_131.2.exe)

Issue #713, opened 2024-01-21
RAEM contest, error text in STX_STRING after ADIF generation


df5en:
v.4.129.2 and v.4.130.0

Here is feedback from the RAEM Contest follow-up. Unexpectedly, there was a small problem with the ADIF file after the contest. I had never seen this effect in any other contests of TR4W. The STX_STRING contained the text **Error generating my exchange**, even though it should contain geographical coordinates. In my case STX_STRING should be a **51N07O**.

This is how I declared the parameter in the CFG file:
`MY QTH=51N07O`

Here is the first data line of the ADIF. All other lines look the same.

Error text in ADIF output:
`<CALL:4>YT3D <BAND:3>80m <QSO_DATE:8>20231224 <TIME_ON:6>000033 <TIME_OFF:6>000033 <RST_SENT:3>599 <RST_RCVD:3>599 <CONTEST_ID:4>RAEM <MODE:2>CW <FREQ:7>3.51512<SRX_STRING:7>1 4N19O <STX_STRING:28>Error generating my exchange <QTH:6>45N19O <SRX:5>00001<STX:5>00001 <OPERATOR:5>DF5EN <EOR>`

Correct, post-processed:
`<CALL:4>YT3D <BAND:3>80m <QSO_DATE:8>20231224 <TIME_ON:6>000033 <TIME_OFF:6>000033 <RST_SENT:3>599 <RST_RCVD:3>599 <CONTEST_ID:4>RAEM <MODE:2>CW <FREQ:7>3.51512<SRX_STRING:7>1 4N19O <STX_STRING:6>51N07O <QTH:6>45N19O <SRX:5>00001<STX:5>00001 <OPERATOR:5>DF5EN <EOR>`

It wasn't a real problem as everything could be fixed with a Search&Replace in the editor. Luckily the value to be replaced was just a constant. Nevertheless, the feedback here is that there is a small problem with ADIF generation (only) in the RAEM contest. Everything else worked perfectly and stably throughout the entire contest.

Vy 73
Dieter

ny4i:
I found the issue but STX_STRING is the whole exchange so it would be the serial number and the coordinates. In my test case, 

[...]

PR #725, opened 2024-02-06
Added code to handle the ADIF export of myexchange

ny4i:
The code to export My Exchange was missing in PostUnit. I also discovered some others that were missing. I updated the code to set STX_STRING to None and log an ERROR message in  the debug log. I will create a new issue to fix the rest. This closes issue #713 .


n4af:
fix [4.131.4](https://www.tr4w.net/4.131/tr4w_setup_4_131.4.exe)

PR #734, opened 2024-02-24
Make external logger selectable

ny4i:
Adds more framework code to support logging to external loggers. This only works with DXKeeper for now, but the shell function is there for ACLog and HRD. Perhaps someone that wants that option will write the code to implement those options. If someone asks for the other loggers, then we can look more into doing it.


ny4i:
I'm not quite what I did to have what look like pull requests from two weeks show up. I suspect I did not pull from the origin but this should all work OK. I made sure I did a pull before I created the branch I used. That is when I saw I received the branches. 

Check a recent change is still there just to be safe.

n4af:
[Fix 4.131.10](https://www.tr4w.net/4.131/tr4w_setup_4_131.10.exe)

PR #738, opened 2024-02-26
Implemented commands to delete a contact

ny4i:
If the contact is deleted, we send a DeleteQSO record. If the QSO is changed, we send a Delete then send a new one.

### DIFF
--- a/tr4w/src/trdos/LOGSUBS2.PAS
+++ b/tr4w/src/trdos/LOGSUBS2.PAS
@@ -487,12 +487,20 @@ begin
     TempRXData.ceQSO_Skiped := False;
     p := TC_YOUHAVERESTOREDTHELASTDELETED;
     LogContactToUDP(TempRXData);
+    if Assigned(externalLogger) then
+       begin
+       externalLogger.LogQSO(TempRXData);
+       end;
   end
   else
   begin
     TempRXData.ceQSO_Skiped := True;
     p := TC_YOUHAVEDELETEDTHELASTLOGENTRY;
     SendDeletedContactToUDP(TempRXData);
+    if Assigned(externalLogger) then
+       begin
+       externalLogger.DeleteQSO(TempRXData);
+       end;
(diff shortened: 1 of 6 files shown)


ny4i:
I believe I found it...

This code in MainUnit is not calling InitializeSerialPorts for thenetwork radios. Yes this functionshould be changed now to CheckAndInitializePorts_ForThisRadio.

``` delphi
menu_reset_radio_ports:
      begin
        logger.info('Resetting radio ports');
        if ActiveRadioPtr.tNetObject <> nil then
        begin
          ActiveRadioPtr.tNetObject.Disconnect;
          ActiveRadioPtr.tNetObject.Connect;
        end
        else if ActiveRadioPtr.tHamLibObject <> nil then
        begin
          ActiveRadioPtr.tHamLibObject.Disconnect;
          ActiveRadioPtr.tHamLibObject.Connect;
        end;
        //else
        //begin // Active radio is using a serial port
          ActiveRadioPtr.CheckAndInitializeSerialPorts_ForThisRadio;
        //end;
        //
        // Handle radio two
        //
        if InActiveRadioPtr.tNetObject <> nil then
        begin
          InActiveRadioPtr.tNetObject.Disconnect;
          InActiveRadioPtr.tNetObject.Connect;
        end
        else if InActiveRadioPtr.tNetObject <> nil then
        begin
          InActiveRadioPtr.tNetObject.Disconnect;
          InActiveRadioPtr.tNetObject.Connect;
        end
        else
        begin
          InActiveRadioPtr.CheckAndInitializeSerialPorts_ForThisRadio;
        end;
        //
      end;
```

As a test, I commented it out and it works now except the band is not set initially until you change the band.

So fix that up and it should be good. Test against real serial ports too as I used a network radio.

n4af:
[Fix 4.131.11](https://www.tr4w.net/4.131/tr4w_setup_4_131.11.exe)

PR #742, opened 2024-03-06
Fixed reset ports and other items in the  description

### LINKED ISSUE #731: URL https://contestonlinescore.com/clubs/?clubfilter=59 does not open when selecting Show Score
The program is doing an open on the value of the field in SCORE READING URL. That only works if the URL has an extension that Windows knows how to process. The better way is to explicitly tell the default browser to open whatever URL is supplied.

For example, if SCORE READING URL is https://contestonlinescore.com/clubs/?clubfilter=59, that URL should be opened in my default browser. As it stands now, the program returns the following error:

image

### LINKED ISSUE #737: After resetting radio ports, the display does not update
Even though the radio is connected (BandUP and BandDN works), the display is not updating. I suspect something in RadioPolling stopped. Perhaps the thread itself needs to be restarted.

ny4i:
Made reset ports work for a networked K4. This is not quite done as the band is not getting set until you change bands I will work onthis in a future change so you do not even need to do the reset on a networked radio. This closes #737

Also updated the OpenURI function so this closes #731 .


n4af:
[fix 4.131.13](https://www.tr4w.net/4.131/tr4w_setup_4_131.13.exe)

PR #761, opened 2024-04-29
Updates for FQP and other QPs

ny4i:
Updated the multipliers for FQP to handle DX multipliers. Updated several QSO parties to add a Cabrillo name where it is different than the string name for each QSO party. Fixed a bug in the PostScores routine which was double counting contacts. This may have been a reported issue for #731


n4af:
`[fix 4.134.2](https://tr4w.net/4.134/tr4w_setup_4_134.2.exe)`

PR #764, opened 2024-05-10
SUBMODE could incorrectly be set where not applicable due to stale data

ny4i:
In the PostUnit where the ADIF file is exported, there are temporary strings used. ExportADIF is just called once and the temp strings were not cleared out after each log record was processed. This has been fixed.


n4af:
[fix 4.135.3](https://www.tr4w.net/4.135/tr4w_setup_4_135.3.exe)

PR #782, opened 2024-08-28
Update uSuperCheckPartialFileUpload.dcu

ny4i:
Updated due to compiling with new version of Indy.


n4af:
Hi Tom -
I do not have this file in my local source. Should I add this  to /src ??

n4af:
never mind-updated.

PR #796, opened 2024-11-27
Added code for 7760 and RegEx maintenance

ny4i:
Added support for the Icom 7760 (based on IC-7610). This has to be tested.

Moved some code into includes to ensure TPerlRegEx is compiled with same version of code as pcre library. This should never change so it can just stay here.

Also put a newly compiled version of SCPFileUpload.dcu to account for using Indy 10. They had to be compiled with same version of Indy.


ny4i:
Howie -did you see this one?

I had an issue with a few libraries like RegEx so let me know if you have any compile issues on master. If so, I can fix those up.

ny4i:
Hi Howie - I have another fix the contest this weekend. Can you integrate this pull request first please?

Then I can generate another one. - Thanks.

n4af:
RR

n4af:
alas, i am headed into town for shopping.
back after lunch.

73, Howie

ny4i:
Here is my config to search for Indy (which is the issue).

image

Here is the file structure in include

image


Let me know what your include folder looks like. 

If I recall, the issue is the order of the  search paths so see if those are right.

Note this is all an issue as the gent from SCP asked me to not publish the code in the clear hence I just send include the dcu and have the SCPUpload project in a private repo that you should also have access to... but I am going to check that right now.

n4af:
It compiles. Perhaps you can test ?
https://tr4w.net/4.140/tr4w_setup_4_140.1.exe

PR #809, opened 2024-12-26
Added IC-7760

ny4i:
Added Icom IC-7760 and also fixed uSuperCheckPartialFileUpl;oad.dcu version again. This version was compiled with Indy 10. As long as another developer is using Indy 10 in their search path, it should match.

If not, email me as we have to find if there is an old version 9 of Indy hiding on the machine that has an error.


ny4i:
Hi. I noticed the entry in the change notes on the website says added IC-7610 support. That should read IC-7760. 73

n4af:
rr done...

n4af:
rr. done. 73, howie

ny4i:
Confirmed. Thanks so much for the update. And again my apologies for making a mess of the repo.

Tom

PR #865, opened 2026-04-10
POTA Nfer parks, 2nd op repeat, RIT SO2R fix

ny4i:
## Summary

- **POTA Nfer label**: QuickDisplay now shows "3fer", "4fer" etc. dynamically based on the actual number of parks logged — was always "2fer"
- **POTA 2nd operator repeat** (Ctrl+T): new Commands menu item pre-fills the exchange with the last logged park references so a second operator can work the same activation without re-typing parks
- **POTA stealth menu**: the two POTA menu items are hidden entirely (not just grayed) when the active contest is not POTA
- **Ctrl+T accelerator**: added to the binary resource table so it dispatches like all other accelerators
- **FlexRadio RIT accumulation fix**: RIT offset was stuck after the first bump because Flex does not reliably echo `rit_freq` back; local offset is now updated optimistically before sending the CAT command
- **SO2R RIT routing fix**: `RITBumpUp`/`RITBumpDown` in `LOGRADIO.PAS` always routed to `Radio1.tNetObject` regardless of which radio was active; now routes to the active radio's net object so Radio 2 (e.g. K4) works correctly

## Test plan

- [x] POTA contest: log a 2fer — verify QuickDisplay shows "2fer park 1 of 2"
- [x] POTA contest: log a 3fer — verify QuickDisplay shows "3fer park 1 of 3", "3fer park 2 of 3"
- [x] Press Ctrl+T after logging a multi-park QSO — verify exchange window pre-fills with last parks
- [x] Open non-POTA contest — verify POTA menu items are not visible
- [x] FlexRadio: press left/right Shift multiple times — verify RIT accumulates beyond first press
- [x] SO2R with K4 as Radio 2: press left/right Shift — verify K4 RIT moves (not Flex/Radio 1)


n4af:
https://tr4w.net/4.146/tr4w_setup_4_146.5.exe

n4af:
Hi Tom -
You might want to announce this on the forum ?

PR #867, opened 2026-04-12
Fix narrow QTH columns on startup; HamLib Direct: WSJT-X minimal polling, KX3 support

### LINKED ISSUE #866: QTH (and other exchange) columns too narrow on startup when log has QSOs
## Summary

Every time TR4W starts with an existing log, exchange columns — specifically QTH — are too narrow to show the data and require manual resizing. This does not happen with an empty log.

## Root cause

`EnsureListViewColumnVisible` (MainUnit.pas:6639) is called at the end of `LoadinLog` when the log has records (line 5379-5380). It uses `LVSCW_AUTOSIZE_USEHEADER`, which on Windows sizes the column to the **header text width** only — not the data content. The header `"QTH"` is 3 characters, but data values such as ARRL sections (`"NEW ENGLAND"`, `"MIDWEST"`, etc.) are much wider, so the column is rendered too narrow.

The same function is also called after every QSO logged (LOGSUBS2.PAS:2681), so the problem persists during the contest as well — but is most noticeable at startup.

## Columns affected

The loop in `EnsureListViewColumnVisible` runs from `logColNumberReceive` to `High(LogColumnsType)`, so the affected columns are:

- QsR (received number)
- DX mult, Zone mult, Prefix mult, Precedence, Check
- Name, Class, **QTH**, Age, Chapter, Power, FOC#, Kids
- Points, Mults, ComputerID, S&P, Dupe, Freq, Op

Band, Date, UTC, QsS (sent number), and Callsign are **not** in this range and always keep their static `Width * ws` values from `ColumnsArray`.

## Proposed fix

Change `LVSCW_AUTOSIZE_USEHEADER` → `LVSCW_AUTOSIZE` in `EnsureListViewColumnVisible`:

```pascal
procedure EnsureListViewColumnVisible(h: HWND);
var
  TempColumn: LogColumnsType;
begin
  for TempColumn := logColNumberReceive to High(LogColumnsType) do
    if ColumnsArray[TempColumn].Enable then
      ListView_SetColumnWidth(h, integer(TempColumn), LVSCW_AUTOSIZE);
end;
```

`LVSCW_AUTOSIZE` sizes each column to the widest data item, which is the expected behavior. The minor trade-off is that `LVSCW_AUTOSIZE_USEHEADER` fills the remaining ListView width for the last column; `LVSCW_AUTOSIZE` does not — but this is a cosmetic difference.

## Files

- `tr4w/src/MainUnit.pas` — `EnsureListViewColumnVisible` (line 6639), `LoadinLog` (line 5379)
- `tr4w/src/trdos/LOGSUBS2.PAS` — line 2681 (post-QSO call)

ny4i:
## Summary

- Fixes n4af/TR4W#866: exchange columns (QTH, etc.) too narrow on startup when log has QSOs
- HamLib Direct: switch to WSJT-X minimal polling to stop front-panel menu interference; add KX3 radio support

---

## Column Width Fix (closes #866)

`EnsureListViewColumnVisible` was using `LVSCW_AUTOSIZE_USEHEADER`, which sizes columns to the header text width only (e.g. "QTH" = 3 chars), ignoring data content like `"NEW ENGLAND"`. Changed to `LVSCW_AUTOSIZE` so columns fit their widest data value.

Added `ColumnAutoSize` flag and `ColumnWidthOverride` array to support persisting user-adjusted column widths across sessions.

Files: `MainUnit.pas`, `uCommctrl.pas`, `tr4w.dpr`, `VC.pas`

---

## HamLib Direct Improvements

**Root cause of front-panel menu dismissal:**
HamLib's Icom driver sends `$07 $D0` (Select Main Band — a SET command, not a query) before every `rig_get_rit`/`rig_get_xit` call because the IC-7610 lacks `RIG_TARGETABLE_RITXIT` capability. This SET command forces the radio to close any open front-panel menu.

**Fix — WSJT-X minimal poll set:**
Adopted the same 6-command poll cycle WSJT-X uses (VFO A freq/mode, PTT, split, VFO B) with no RIT/XIT in the fast heartbeat. RIT/XIT moved to a separate 5-second slow poll (`SendRITXITPoll`), reducing `$07 $D0` interference to once every 5 seconds — imperceptible in normal operation.

**Accurate RIT/XIT state:**
Uses `rig_get_func(RIG_FUNC_RIT/XIT)` for on/off state, separate from `rig_get_rit`/`rig_get_xit` offset values. Fixes false "RIT active" display when IC-7610 stores a non-zero offset with RIT disabled.

**Async transceive callbacks:**
Registered via `rig_set_trn(RIG_TRN_RIG)` + `rig_set_freq_callback` etc. for fast response when the radio does fire them. On serial HamLib backends callbacks are unreliable, so heartbeat polling remains primary.

**HamLib trace logging:**
`HAMLIB TRACE = TRUE` in config redirects HamLib's internal debug output to `hamlib_trace.log` for diagnostics.

**HamLib warning in CAT dialog:**
When HamLib is selected for a radio with native TR4W support, a warning explains the RIT/XIT limitation.

**KX3 radio support:**
Added KX3 between K3 and K4 in `LOGRADIO.PAS` and `VC.pas` (Kenwood protocol, 38400 baud, HamLib ID 2045).

Files: `uRadioHamLibDirect.pas`, `uHamLibDirect.pas`, `uRadioPolling.pas`, `uCAT.pas`, `uCFG.pas`, `VC.pas`, `LOGRADIO.PAS`

---

## Test plan

- [ ] Start TR4W with an existing log containing QSOs — verify QTH and exchange columns are wide enough to show data without manual resize
- [ ] Use HamLib Direct with IC-7610 or IC-7760 — verify pressing mode button on radio does not dismiss front-panel display
- [ ] Confirm RIT/XIT on/off state displays correctly in TR4W when radio RIT is enabled/disabled
- [ ] Verify RIT/XIT offset values update within ~5 seconds of change on radio
- [ ] Test KX3 radio selection and CAT connectivity
- [ ] Enable `HAMLIB TRACE = TRUE`, confirm `hamlib_trace.log` is created with output
- [ ] Verify HamLib warning appears when selecting HamLib for a natively-supported radio (e.g. IC-7610)

### DIFF
--- a/tr4w/src/MainUnit.pas
+++ b/tr4w/src/MainUnit.pas
@@ -322,6 +322,7 @@ procedure ProcessCommandLine;
 procedure PutCallToCallWindow(Call: CallString);
 procedure SetColumnsWidth;
 procedure EnsureListViewColumnVisible(h: HWND);
+procedure SaveColumnWidthToConfig(ColIndex: Integer; NewWidth: Integer);
 procedure ExecuteConfigurationFile(f: ShortString);
 procedure CheckEditableWindowHeight;
 function CheckCommandInCallsignWindow: boolean;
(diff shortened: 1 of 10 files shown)


n4af:
https://tr4w.net/4.146/tr4w_setup_4_146.6.exe

Issue #688, opened 2023-09-01
Lock the bandmap window while updating to prevent flashing


ny4i:
During frequent updates, the bandmap flashes from updating so frequently. One possible solution is to lock the window with the Win32 LockWindowUpdate function. 

I do not recall if I tried the this before.

Using this ticket to track it.

https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-lockwindowupdate

ny4i:
The LockWindowUpdaes suggested to send the message [WM_SETREDRAW](https://learn.microsoft.com/en-us/windows/desktop/gdi/wm-setredraw) instead. I tried that and while it made some difference, the BM still flashes.

One thing I learned is it appears the telnet code only uses a thread for the connection. I thought I must be reading that wrong, but when I added a sleep (250) to sleep 250 milliseconds, the whole UI slowed down. My thought was to limit how often the update updates but the telnet thread should be in a thread instead. 

Another thought is to set a TTimer to fire every 250 milliseconds and do the update. 

Even without adding full threading for telnet (which really should happen), that can be tested by setting BMDELAY to 250.

n4af:
alas-
the spot must sync to the display. else, clicking on a b.m. entry displays
the wrong call entry/frequency..
likewise, delaying spot processing results in lost spots.

ny4i:
Normally, the way to handle that is to store an object with the list box entry which is the spot data. That way clicking on any spot will always give the correct info. I’m not sure how to do that yet with Win32 but I’ll look. TomPrincipal Solutions ArchitectBetter Software Solutions, Inc. 727-437-2771On Sep 1, 2023, at 12:59 PM, Howie Hoyt ***@***.***> wrote:﻿
alas-

the spot must sync to the display. else, clicking on a b.m. entry displays

the wrong call entry/frequency..

likewise, delaying spot processing results in lost spots.



On Fri, Sep 1, 2023 at 12:14 PM Tom Schaefer ***@***.***>

wrote:

ny4i:
The above is wrong. The Telnet code is definitely in a thread but the part where it calls DisplayBandMap is not so more checking is required.

ny4i:
[...]

PR #869, opened 2026-04-12
Band map flicker reduction, spot deletion fix, radio logging cleanup

ny4i:
## Summary

- **Band map flicker on spot arrival** (#688): 250ms coalesced refresh timer, `WS_EX_COMPOSITED`, `RDW_NOERASE` on redraws, remove duplicate `DisplayBandMap` on startup — flicker greatly reduced
- **Band map flicker on VFO change**: `ProcessFilteredStatus` was calling `DisplayBandMap` directly on every poll cycle; now sets `BandMapNeedsRefresh` so VFO movement coalesces through the same 250ms timer
- **`BandMapNeedsRefresh` relocated**: moved from `uTelnet` to `LogWind` alongside other `BandMap*` globals, eliminating circular dependency risk with `uRadioPolling`
- **Spot deletion**: `DeleteSpotFromBandmap` now calls `DisplayBandMap` immediately; previously deleted spots stayed visible until the next spot arrival
- **K4 serial VFO responsiveness**: `pollingInterval` now follows `FREQUENCY POLL RATE` config setting (default 10ms) instead of hardcoded 1000ms — VFO display updates now match K3 speed
- **Clean shutdown (K3/K4 serial)**: `pKenwood2` and `pKenwoodNew` now check `PollingStopRequested` at the outer loop and inner byte-wait loop — K3/K4 serial polling threads exit in ~20ms on quit instead of timing out after 3000ms
- **Radio logging**: Demote ~25 per-packet/sub-step `logger.Info` calls to `logger.Debug` in `uRadioIcomBase.pas` and `uIcomNetworkTransport.pas`

## Notable: 250ms timer (needs N4AF review)

The coalesced refresh timer is a behavioral change — spot arrivals and VFO changes no longer trigger an immediate repaint; the display updates at most once every 250ms. See comment on #688 for full details. N4AF approval requested before merge.

## Test plan

- [ ] Start TR4W with a saved `BandMap.bin` — verify clean startup, no flash
- [ ] Connect to DX cluster with high spot rate — verify band map updates smoothly without constant flicker
- [ ] Tune VFO — verify band map cursor moves without flashing
- [ ] Right-click band map → Delete a spot — verify it disappears immediately
- [ ] Quit TR4W with K3 or K4 serial connected — verify shutdown is near-instant (was ~4 seconds)
- [ ] Spot click tunes correct radio (regression check for `$26 $01` VFO B mode write)
- [ ] Spot comment containing "Pup" or similar does not trigger false split (UP parser fix regression)
- [ ] Radio log at INFO level no longer shows per-packet CI-V responses

### DIFF
--- a/tr4w/src/MainUnit.pas
+++ b/tr4w/src/MainUnit.pas
@@ -263,6 +263,8 @@ procedure PTTOffWhenStopWAV(uTimerID, uMessage: UINT; dwUser, dw1, dw2: DWORD)
   stdcall;
 procedure OneSecTimerProc(uTimerID, uMessage: UINT; dwUser, dw1, dw2: DWORD)
   stdcall;
+procedure BandMapRefreshTimerProc(uTimerID, uMessage: UINT; dwUser, dw1, dw2: DWORD)
+  stdcall;
 
 procedure SaveTR4WPOSFILE;
 procedure LoadTR4WPOSFILE;
@@ -1570,6 +1572,16 @@ begin
   // Windows.SetWindowText(tr4whandle, inttopchar({GetHeapStatus.TotalFree}AllocMemSize));
  // Windows.SetWindowText(InsertWindowHandle, inttopchar(FreeMemCount));
 {$IFEND}
+end;
+
+procedure BandMapRefreshTimerProc(uTimerID, uMessage: UINT; dwUser, dw1, dw2: DWORD)
+  stdcall;
+begin
+  if BandMapNeedsRefresh then
+     begin
+     BandMapNeedsRefresh := False;
+     DisplayBandMap;
+     end;
 end;
 
 procedure PTTOffWhenStopWAV(uTimerID, uMessage: UINT; dwUser, dw1, dw2: DWORD)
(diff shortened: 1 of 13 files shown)


n4af:
https://tr4w.net/4.146/tr4w_setup_4_146.7.exe
not propagating to remote master until more testing

ny4i:
I found a few issues related to the bandmap refreshing in multiple places while testing ExchangeRadios. I am fixing that and will update the PR.

n4af:
Fb Tom

Tks, h

ny4i:
Pull the latest pr code and build please. This fixes a screen flicker on radio 2 updates

n4af:
https://tr4w.net/4.146/tr4w_setup_4_146.8.exe

n4af:
Hi Tom -

I do see one problem. if i click on spot 'w4abc/m' it loads correct.
but if I then select r1zn, it will load into call window with residual junk
'r1znc/m'
i think we had this bug before where the call field was not being cleared.

tks, howie

n4af:
This only occurs with the ALT-D window.
the main call window clears ok.

Issue #872, opened 2026-04-14
Alt-D fails to clear on entry of new spot


n4af:
bandmap set to 'BANDMAP ALL BANDS' and  'QSY INACTIVE RADIO'
TUNE ALT-D ENABLE set to TRUE.

Click on inactive band spot HB0/PA3CNO' and it is correctly loaded into alt-d window.

Now click on inactive band spot  'DF9II' this loads alt-d window with residual > length DF9II, eg
'DF9IIA3CNO' 
Pressing space bar correctly loads DF9II, but it appears CLEARALTD is not clearing the
alt-d window before loading next call.

n4af:
noted using 4.146.8

n4af:
<img width="1440" height="900" alt="Image" src="https://github.com/user-attachments/assets/befafd38-8f59-4293-974d-670de20df9c8" />

n4af:
trace does not show the problem.
I will pursue later.

ny4i:
## Root cause analysis

**File:** `uBandmap.pas` — double-click handler for QSY INACTIVE RADIO (`LBN_DBLCLK`, added in v4.135.1 / commit `8ce3ce8`)

**The bug:** `DupeInfoCall` is a `Str20` (Pascal ShortString). When a new callsign is assigned to it, only the length byte and the new characters are written — the remaining bytes in the buffer are *not* zeroed. `DupeCheckOnInactiveRadio` then passes `@DupeInfoCall[1]` as a raw PChar to `wsprintf`, which reads until it finds a `#0` byte. If the new callsign is shorter than the previous one, the stale bytes from the old call are read as part of the new string.

**Example:** `HB0/PA3CNO` (10 chars) → `DF9II` (5 chars). After assigning `DF9II`, bytes 6–10 still contain `A3CNO`. `wsprintf` reads `DF9IIA3CNO` before finding `#0` at position 11.

**The established fix pattern** already exists in `uSpots.pas` `TuneDupeCheck` (lines 711–714):
```pascal
tClearDupeInfoCall;   // ZeroMemory on the full Str20 buffer
ClearAltD;
DupeInfoCall := FList^[Index2].FCall;
DupeCheckOnInactiveRadio(True);
```

**The omission:** The bandmap double-click path in `uBandmap.pas` was added in v4.135.1 without applying this pattern:
```pascal
// before fix:
dupeinfocall := spot.fcall;
dupecheckoninactiveradio(true);
```

[...]

PR #873, opened 2026-04-14
Fix Alt-D residue when clicking shorter callsign (issue #872)

ny4i:
## Summary

- **Root cause:** In `uBandmap.pas`, the double-click path for QSY INACTIVE RADIO assigned `DupeInfoCall` without zeroing the buffer first. `DupeCheckOnInactiveRadio` passes `@DupeInfoCall[1]` as a raw PChar to `wsprintf`, which reads until it finds `#0`. When the new callsign is shorter than the previous one, stale bytes from the old call are read as part of the new string (e.g. clicking `DF9II` after `HB0/PA3CNO` shows `DF9IIA3CNO`).
- **The missing call:** `tClearDupeInfoCall` (which does `ZeroMemory` on the full `Str20` buffer) was not called before the assignment. The correct pattern is established in `uSpots.pas` `TuneDupeCheck` (lines 711-714) and was simply not applied when this code path was added in v4.135.1 (commit 8ce3ce8).
- **Fix:** Add `tClearDupeInfoCall` + `ClearAltD` before `dupeinfocall := spot.fcall` in `uBandmap.pas`, matching the pattern from `uSpots.pas`. Also added `uDupesheet` to `uBandmap`'s implementation uses to make `ClearAltD` visible.

## Repro sequence

1. SO2R mode, `QSY INACTIVE RADIO = TRUE`, `TUNE ALT-D ENABLE = TRUE`, both radios reporting frequency
2. Double-click a **long** callsign in the bandmap (e.g. `HB0/PA3CNO`) — displays correctly
3. Double-click a **shorter** callsign (e.g. `DF9II`) — Alt-D window shows `DF9IIA3CNO` (residue from step 2)

## Test plan

- [ ] Double-click long callsign in bandmap with QSY INACTIVE RADIO enabled — Alt-D shows correct call
- [ ] Double-click shorter callsign — Alt-D shows only the new call, no residue
- [ ] Verify space bar still logs the correct call
- [ ] Verify single-radio mode bandmap click unaffected

### DIFF
--- a/tr4w/src/uBandmap.pas
+++ b/tr4w/src/uBandmap.pas
@@ -116,7 +116,7 @@ var
   BandMapDisplayGhz: boolean;
 
 implementation
-uses MainUnit;
+uses MainUnit, uDupesheet;
 
 function BandmapDlgProc(hwnddlg: HWND; Msg: UINT; wParam: wParam; lParam:
   lParam): BOOL; stdcall;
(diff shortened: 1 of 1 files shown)


ny4i:
Please see if this one fixes the issue. I could not figure ouot how to duplicate it.

n4af:
genius. !    
https://tr4w.net/4.146/tr4w_setup_4_146.9.exe

PR #886, opened 2026-05-09
Issue #885: one Enter press = N QSOs for multi-county and multi-park

### LINKED ISSUE #885: rewrite support for county lines
For stateside QSO parties allow single line entry for stations bridging two county lines.
1) You must be in S&P mode
2) Exchange must be entered as  'county abbreviation1/countyabbreviation2'.
For example, in FL QP, enter exchange as 'DAD/BRO' of the station indicates they are on the county line
between Dade and Broward.

From the above entry, TR4W will generate two line entries using the same callsign but one line county would be 'DAD' and the second log entry for county  ' BRO'.
Because the second line in the log is assigned to the same call, TR4W will indicate dupe points 0, but the cabrillo generates ok- as always leaving assignment of 'DUPE' to the contest sponsor.

If a station is giving out > 2 county lines, you may use the above for the first two add subsequent lines for any additional counties.

ny4i:
## Summary

Replaces the prior PostMessage refill cycle (POTA Nfer "press Enter again to start 2fer") with inline drain of additional QSO records right after the first `LogContact`. Same shape applied to state-QP county-line entry (`DAL/BAY`, `DAL BAY`, `599 DAL/BAY`, and any combination thereof) so a single keystroke produces N log records.

Closes #885 by landing a more thorough implementation than the earlier 4.146.16 attempt:
- Works in **both S&P and CQ modes** (previous version was S&P only)
- Handles **slash form** (`DAL/BAY`) **and space form** (`DAL BAY`)
- Handles **3+ counties** via repeated `/`
- Removes the press-Enter-again refill cycle for **POTA Nfer** as a side benefit

## Architecture

**Parser (`LOGSTUFF.PAS::ProcessRSTAndDomesticQTHExchange`)**
- Maps `/` to space so slash and space forms travel one code path
- Walks every token, probes each against the domestic-mults table with a scratch `ContestExchange`, collects every valid QTH in user-typed order
- First valid QTH becomes `RXData.QTHString` for THIS QSO; the rest are pushed onto a new `uPendingCounties` queue

**Drain (`MainUnit.pas::DrainPendingMultiQSORefs`)**
- Called once from `TryLogContact` right after the first `LogContact`
- Pops both queues (POTA parks, then counties) and writes one extra QSO per ref using a stack-local `ContestExchange` copy
- Refreshes `DomMultQTH`/`DomesticQTH` per county via `FoundDomesticQTH` so each record carries the correct multiplier

**Removed (no longer needed):**
- `WM_POTA_NEXT_PARK` constant + `tr4w.dpr` dispatch case
- `HandlePOTANextPark` proc (interface decl + 56-line implementation)
- `SetPendingContactInfo` / `GetPendingCall` / `GetPendingRST` / `SetTotalParks` / `GetTotalParks` helpers and their state vars in `uPOTAParks`
- The `IsFresh` / `GetPendingCall` comparison in `ProcessRSTAndPOTAPark` (queue is now always rebuilt from the current exchange)
- The slash-detection block in `ReturnInSAPOpMode` and its `goto loop` follow-up (parser handles the split now; works in CQ mode too)

## Why `DrainPendingMultiQSORefs` is its own procedure

Inlining the drain loops directly into `TryLogContact` pushes the function past a Delphi 7 codegen threshold where the compiler pins `@ReceivedData` into ESI and reuses ESI for every subsequent `ReceivedData` access. At least one code path inside the new call chain does not preserve ESI, so the post-loop `ShowStationInformation(@ReceivedData.Callsign)` gets a stale base register and AVs. Confirmed via debugger: `LEA EAX,[ESI+\$2C]` with ESI pointing into the binary's read-only constant pool instead of the data segment.

Keeping the drain in its own procedure leaves `TryLogContact`'s register allocation identical to master's working pattern.

## Behaviour

| Input | Result |
|---|---|
| `DAL` | 1 QSO, QTH=DAL (single county, S&P or CQ) |
| `DAL/BAY` | 2 QSOs, QTH=DAL then QTH=BAY (slash form) |
| `DAL BAY` | 2 QSOs, QTH=DAL then QTH=BAY (space form) |
| `599 DAL/BAY` | 2 QSOs, RST=599 carried forward |
| `DAL/BAY/CIT` | 3 QSOs, one per county |
| `DAL/XXX` (XXX invalid) | 1 QSO, QTH=DAL (falls through to existing error path) |
| `57 US-1234 US-5678` (POTA) | 2 QSOs, one per park |

The second/Nth log of the same call is flagged dupe with 0 points by `TryLogContact`'s normal dupe handling; Cabrillo writes both lines so the contest sponsor handles the dupe accounting. This matches the behaviour spelled out in the issue.

Rover slash-in-call (`KG1S/PIN`) lives in a different code path and is untouched.

## Test plan

- [x] Single-county FQP: `W4IU` + `PIN` → 1 QSO logged
- [x] Multi-county slash: `W4IU` + `HIL/PIN` → 2 QSOs logged
- [x] Multi-county space: `W4IU` + `HIL PIN` → 2 QSOs logged
- [x] POTA 2fer: `57 US-1234 US-5678` → 2 QSOs logged in one Enter
- [ ] CQ mode multi-county (additional confirmation)
- [ ] Rover call form (`KG1S/PIN` in call window) still pre-fills exchange and logs as one QSO with stripped call on Cabrillo output
- [ ] Ctrl+T POTA Repeat-for-2nd-Op still works (uses unrelated `GetLastPOTAExchange`)
- [x] Cabrillo output for a multi-county QSO contains both lines

### DIFF
--- a/tr4w/build/full.nsi
+++ b/tr4w/build/full.nsi
@@ -1,4 +1,4 @@
-!define TR4WVERSION    '146.14'
+!define TR4WVERSION    '146.15'
 !define TR4WINSTFOLDER 'Software\TR4W'
 !define TR4WDRVREG     'SYSTEM\CurrentControlSet\Services\TR4WIO'
 
--- a/tr4w/build/full.nsi
+++ b/tr4w/build/full.nsi
@@ -1,11 +1,12 @@
-!define TR4WVERSION    '146.16'
+!define TR4WVERSION    '147.00'
 !define TR4WINSTFOLDER 'Software\TR4W'
 !define TR4WDRVREG     'SYSTEM\CurrentControlSet\Services\TR4WIO'
 
 ;!define MMTTYMODE  'mmtty'
 ;!define TR4WLANG    'rus'
-;!define TR4WLANG    'mng'
 ;!define TR4WLANG    'cze'
+!define TR4WLANG    'mng'
+
 ;!define TR4WLANG    'ser'
 ;!define TR4WLANG    'rom'
 ;!define TR4WLANG    'ger'
--- a/tr4w/build/full.nsi
+++ b/tr4w/build/full.nsi
@@ -1,4 +1,4 @@
-!define TR4WVERSION    '146.15'
+!define TR4WVERSION    '146.16'
 !define TR4WINSTFOLDER 'Software\TR4W'
 !define TR4WDRVREG     'SYSTEM\CurrentControlSet\Services\TR4WIO'
 
--- a/tr4w/src/MainUnit.pas
+++ b/tr4w/src/MainUnit.pas
@@ -410,7 +410,6 @@ procedure tCleareCallWindow;
 procedure tCleareExchangeWindow;
 procedure tSetExchWindInitExchangeEntry;
 procedure HandleRepeatPOTAParks;
-procedure HandlePOTANextPark;
 procedure tListBoxClientAlign(Parent: HWND);
 //function AddCallsignAndExchangeToInitialExchangesList(Call: CallString; InitialExchangeString: CallString): boolean;
 //function FindStringInInitCallsignListBox(s: CallString; var Index: integer): boolean;
@@ -521,6 +520,7 @@ uses
   // Country9,
   FCONTEST,
   uPOTAParks,
+  uPendingCounties,
   uCTYUpdate,
   Types;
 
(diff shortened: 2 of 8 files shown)


n4af:
https://tr4w.net/4.147/tr4w_setup_4_147.01.exe

ny4i:
I am working on the merge conflict.

ny4i:
OK. Please merge this PR.

n4af:
dunno. i need spaces before/after slash for it to work ?

Issue #889, opened 2026-05-10
For multi-line county QP, set SRX_STRING to just one county


ny4i:
For multi-line counties or parks, the SRX string is set to the multi string (e.g., PIN/HIL/MAN). Each QSO should have just the county in question. so if I enter 57 PIN/HIL, the first contact SRX_STRING should be 57 PIN and th esecond QSO 57 HIL.

Issue #877, opened 2026-04-15
When running stations during POTA, the parser does not accept free-form info like state or name


ny4i:
The function ProcessRSTAndPOTAPark in LOGSTUFF.pas:4580 is limiting itself to expecting just the following in the exchange:
// Parses a POTA exchange of the form:
//   RST PARK              e.g. "57 US-0663"
//   RST PARK PARK         e.g. "57 US-0663 US-1234"   (2fer)
//   RST PARK PARK PARK    e.g. "57 US-0663 US-1234 US-5678"  (3fer, etc.)
//   PARK                  e.g. "US-0663"              (no RST)

As POTA is not a formal contest with a set exchange, this should accept just a state or a name. The parser should run the functions LooksLikeAState, LooksLikeAGrid, LooksLikeASection, LooksLikeRST against the tokens in the exchange string and place the parts in the proper place in ContestExchange.

I should able to enter just a state, or a state and a park, or just an RST, and a Name.  If you wanted to gild the lily, you could write a function called LooksLikeAName and search the name database (look at LOGSCP.pas CallDatabase.GetName for an example of grabbing the names out of the SCP file). That would need instrumenting to show timings to ensure the file is loaded into a data structure that provides the best possible search (binary search on a sported list of names from the database)? Just in case the callsign entered is already inthe SCP, a call to GetName for the callsign would be useful), but otherwise I am suggesting the names inthe SCP database could be a list of known contester names. If the string in the exchange does not match a name, then you could just leave it in the SRX_STRING and not worry about putting anywhere else.

ny4i:
## Clarification & Claude Prompt Notes

This issue requests that the `ProcessRSTAndPOTAPark` function in `LOGSTUFF.pas` (around line 4580) be reworked into a more flexible, token-based exchange parser for POTA operations.

---

### Current Behavior
The parser only accepts structured POTA exchanges in rigid forms:
- `RST PARK` — e.g. `57 US-0663`
- `RST PARK PARK` — e.g. `57 US-0663 US-1234`
- `RST PARK PARK PARK` — e.g. `57 US-0663 US-1234 US-5678`
- `PARK` — e.g. `US-0663` (no RST)

### Desired Behavior
Since POTA is not a formal contest with a fixed exchange, the parser should accept **free-form exchange tokens** and classify each token by type. Valid token types include:
- **RST** — identified by `LooksLikeRST`
- **POTA Park reference** — e.g. `US-0663`, identified by existing POTA park pattern matching
[...]

Issue #770, opened 2024-06-23
Add the operator to the Network window display


ny4i:
It would be nice to see the call of the active operator at each station on the network

ny4i:
## Implementation Analysis

### How the server handles this

The server's `WM_SOCK_NET_RX` handler for `NET_STATIONSTATUS_ID` is a blind relay — it forwards raw bytes using `SizeOf(TStationState)` without inspecting `ssType`. So adding the new operator field to `TStationState` and a new `sstOperator` status type requires **no code changes to tr4wserver** — just a recompile against the updated definition.

One pre-existing limitation (not introduced by this feature): the server does not cache per-client station state to replay to newly-joining clients, so a late-joining station won't see operator names until an update is sent. This is the same behavior as the existing computer name and callsign fields today.

### Change set

| File | Change |
|------|--------|
| `tr4w/src/VC.pas` | Add `sstOperator` to `StationStatusType`; add `ssOperator: OperatorType` to `TStationState` |
| `tr4w/src/uNet.pas` | Add "Op" column to `NetColumnsArray`; handle `sstOperator` in `SendStationStatus` and `DisplayClientStatus` |
| `tr4w/src/MainUnit.pas` | Call `SendStationStatus(sstOperator)` where `CurrentOperator` changes |
| `tr4w/tr4wserver/tr4wserver.dpr` | Recompile only — no code changes needed |

Issue #674, opened 2023-07-16
Add ID field to UDP Broadcasts


ny4i:

ny4i:
This was never actually committed to master. Working on in now

PR #890, opened 2026-05-10
May 2026 — accumulated fixes, features, and tests

ny4i:
Bundle of accumulated fixes, features, and tests that have been tested individually but were waiting for a PR window. Nine commits, each with a self-contained message for traceability.

### Bug fixes

- **`0bbb649`** Issue #674 — `uNewContest`: delete `.TRW` when overwriting an existing contest. Without this, opening the new contest hits a fatal "log file is not the correct size" because the leftover `.TRW` from the old contest is still around.
- **`88d10a3`** Issue #26 — `uGetScores`: replace raw WinSock2 POST with `TIdHTTP` + `TIdSSLIOHandlerSocketOpenSSL` (TLS 1.2). Default URL set to `https://post.contestonlinescore.com/post/` (the bare host returns 404). SSL handler attached only when URL starts with `https://` so operator-configured `http://` custom URLs remain functional. Includes URL in diagnostic log lines for future troubleshooting.
- **`a19a0d5`** `uWinKey` — replace inline-asm `wsprintf` for the COM-port-open error with `Format` + `PChar(msg)`. Old asm pushed `eax` after `SysErrorMessage` assuming PChar return, but Delphi 7 returns `AnsiString` via a hidden var-parameter — the result was garbage like `Winkeyer port COM5: xlk ,ÿñ"nS`.

### Features

- **`6708783`** Issue #770 — show active operator in the network window. Adds `sstOperator` status type and `ssOperator` field to `TStationState`; broadcasts on every operator change; new "Op" column in the network listview. TR4WServer requires only a recompile (its blind-relay path handles the new packet automatically).
- **`b417054`** Issue #47 — Programmable Messages dialog: insert command at cursor on double-click. Operators can now double-click an entry in the "List of Commands" list to insert it at the cursor position in the message text field. Cursor position captured via the existing `EN_KILLFOCUS SelPos[102]` mechanism.
- **`1ff9234`** Issue #889 — per-QSO `SRX_STRING` for multi-county / multi-park entries. When the operator enters `57 PIN/HIL`, each QSO record now carries its own `ExchString` (`57 PIN` for QSO 1, `57 HIL` for QSO 2) rather than the combined original input. Affects ADIF export (`SRX_STRING`).
- **`e528024`** State-QP rover (`KG1S/PIN`) — when the call has a `/COUNTY` suffix and the suffix validates as a domestic QTH, strip the suffix off the call and pre-fill the exchange with the county. The QSO is then logged with the bare call and the county, so Cabrillo/ADIF output is correct without any output-side stripping. Triggers only for QP exchange types and only when the suffix is a real county (so `/M`, `/P`, `/4` etc. fall through unchanged).

### Housekeeping

- **`f50c636`** `uNet` — log connect retries only on state change. The 5-second TCP retry loop was emitting four DEBUG lines per attempt; now logs once when the loop starts, once on success, once on first failure, and silently otherwise. Thread-exit message demoted to TRACE.

### Tests

- **`ff88e00`** `tests/unit/uTestUtilsText.pas` — 27 unit-test cases for `StringIsAllNumbers`, `StringIsAllNumbersOrSpaces`, `StringIsAllNumbersOrDecimal`, `StringIsAllAlphanumericOrDash` (including the `bNoCase` parameter added in #877), and `StringWithFirstWordDeleted`. `FullBuild.ps1` test compile path extended to include `src/` so `VC.pas` (and `Log4D.dcu`) resolve correctly.

## Test plan

Each fix has been tested individually during development. Suggested final test sequence:

- [ ] Issue #674: start a new contest over an existing one — verify no "log file size" fatal
- [ ] Issue #26: post a score, watch the log for the URL line and a successful HTTP 200
- [ ] Network log noise: start TR4W with multi-op enabled but TR4WServer NOT running — verify only one DEBUG line per state transition (not four per retry)
- [ ] Issue #770: change operators on multiple stations — verify "Op" column in network window updates
- [ ] Issue #47: open Programmable Messages, position cursor, double-click an item — verify insertion at cursor
- [ ] WinKey ABI: disconnect WinKeyer, retry — verify error dialog shows real Windows error string, not garbage
- [ ] Issue #889: log a multi-county QSO (`DAL/PIN`), export ADIF — verify each QSO has its own SRX_STRING
- [ ] Rover (`KG1S/PIN`): enter in a state QP — verify call window shows `KG1S` after the slash is detected, exchange auto-fills with `PIN`, Cabrillo line has bare call + correct county

## Known follow-up

Documented as TODOs for a future PR (not in this one):
- QP rover dupe/scoring behaviour: with bare-call logging, repeated QSOs from same call in different counties are flagged dupe by TR4W's standard logic; contest sponsor generally awards them, but TR4W's local score is conservative.

### DIFF
--- a/tr4w/FullBuild.ps1
+++ b/tr4w/FullBuild.ps1
@@ -20,7 +20,8 @@ $TEST_DCU_DIR = "C:\Temp\tr4w-test"
 New-Item -ItemType Directory -Force -Path $TEST_DCU_DIR | Out-Null
 
 Push-Location $TEST_DIR
-& $DCC32 $TEST_DPR -`$D+ -`$L+ -`$Y+ "-N$TEST_DCU_DIR" "/E$TEST_DIR"
+# /U includes src\ so VC.pas (and Log4D.dcu within it) resolve correctly
+& $DCC32 $TEST_DPR -`$D+ -`$L+ -`$Y+ "-N$TEST_DCU_DIR" "/E$TEST_DIR" "/UC:\TR4W\tr4w\src"
 $testBuildResult = $LASTEXITCODE
 Pop-Location
 
--- a/tr4w/src/MainUnit.pas
+++ b/tr4w/src/MainUnit.pas
@@ -591,6 +591,64 @@ begin
     end;
 end;
 
+// Issue: state-QP rover slash-in-call (e.g. "KG1S/PIN").  When the
+// operator types a call with a /COUNTY suffix, treat the suffix as the
+// rover's current location: strip the /COUNTY off the call and pre-fill
+// the exchange field with the county.  Net effect — the QSO is logged
+// with the bare call (KG1S) and county PIN, and the Cabrillo / ADIF
+// output is correct without any output-side stripping.
+//
+// Only fires when:
+//   - active exchange is a state-QP type
+//   - the call contains exactly one '/'
+//   - the suffix validates as a domestic QTH (so /M, /P, /4 etc. are left
+//     alone for other code paths to handle)
+//   - the exchange field is empty (operator hasn't typed something else)
+//
+// Note: this means subsequent rover QSOs from the same callsign in
+// different counties will be flagged as dupes (same logged callsign).
+// Address scoring/dupe behaviour for QP rovers in a follow-up issue.
+
+procedure ApplyRoverSlashInCall;
+var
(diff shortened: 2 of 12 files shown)


ny4i:
Closing — opening prematurely. Will reopen after testing.

Issue #887, opened 2026-05-09
Refactor ADIF import/export to expose in-memory string entry points (enable round-trip unit tests)


ny4i:
## Goal

Make TR4W's ADIF import and export testable by splitting the parser and emitter from file I/O. After this change, a unit test can build QSOs in code, export them to a string, parse the string back, and assert equivalence — no file system, no network.

## Problem

Today TR4W's ADIF import/export is coupled to file streams. That makes a round-trip unit test impossible without touching disk, which is slow, leaks fixture files, and is awkward in CI.

This is also a prerequisite for the N1MM-style log receiver follow-up issue, which needs to feed received ADIF strings into the import path without writing them to disk first.

### Import side
- [ ] Expose `ImportADIFFromString(s: string): TQSORecordList` (or equivalent — match existing TR4W naming).
- [ ] Existing file-based import becomes a two-line wrapper: read the file into a string, call `ImportADIFFromString`.
- [ ] Parser accepts standard ADIF tag-length-value (`<TAG:n>value`) with `<EOR>` / `<EOH>` terminators.
- [ ] No file handles, no streams below the wrapper layer.

### Export side
- [ ] Expose `ExportADIFToString(qsos: TQSORecordList): string` (or equivalent).
- [ ] Existing file-based export becomes a wrapper: call `ExportADIFToString`, write the result to disk.
- [ ] No file handles, no streams below the wrapper layer.

### QSO builder
- [ ] A factory for constructing test QSOs in code, e.g. `MakeQSO(call, band, mode, ...)`. May already exist informally — surface it as a deliberate test helper.

### Round-trip unit test
- [ ] Build N test QSOs covering: every supported mode, every band, the exchange field types TR4W cares about, and edge cases (UTF-8 in NAME/QTH/COMMENT, empty optional fields, max-length call, FREQ precision).
- [ ] `exported := ExportADIFToString(qsos)`
- [ ] `parsed := ImportADIFFromString(exported)`
- [ ] Assert `parsed` equals `qsos` field-by-field. Failures must name the diverging field and QSO index.
- [ ] Runs as part of the existing TR4W test suite (or stands up a new one if none exists).

## Acceptance criteria

- [ ] String-in / string-out functions exist and are public.
- [ ] File-based import/export uses them.
- [ ] Round-trip unit test exists, passes, and covers the field set above.
- [ ] No regression in file-based import/export against existing fixtures.

## Out of scope

- Network-based ADIF reception (separate follow-up issue).
[...]

PR #895, opened 2026-05-11
Stop treating multi-county exchanges as callsign updates (LooksLikeACallSign)

ny4i:
## Summary

In S&P mode with `CALLSIGN UPDATE ENABLE = TRUE`, typing a multi-county exchange like `PIN/HIL` into the exchange window caused TR4W to strip the exchange out as a "corrected callsign", leaving nothing for the exchange parser.  The QSO would not log.

`LooksLikeACallSign` in `trdos/LOGSTUFF.PAS` was returning True for any token whose first interior character was `/`, without requiring any letter↔digit transition first.  Real callsigns always contain digit(s) between prefix-letters and suffix-letters (1-2 letters, digit, 1-3 letters), so they accumulate `nChanges >= 2` by the time the loop reaches the slash.  County-line tokens like `PIN/HIL` have all-letter halves and `nChanges = 0`.

## Fix

Delete the early-True-at-interior-slash branch.  Keep the early-False-at-edge-slash branch (a `/` at position 1 or at the last position is still not a valid callsign shape).  Let the final `nChanges >= 2` check decide.

## Behavior matrix

| Token | Before | After | Correct? |
|---|---|---|---|
| `PIN/HIL` | True | **False** | ✓ (county-line) |
| `PIN`     | False | False | ✓ |
| `KG1S/MON` | True | True | ✓ (rover call) |
| `G/W4XYZ` | True | True | ✓ (single-letter portable prefix) |
| `K/W1AW`  | True | True | ✓ |
| `NY4I`    | True | True | ✓ |
| `NY4A→NY4I` correction | works | works | ✓ |
| `PIN/HIL NY4I` (county-line + correction combined) | broken (PIN/HIL stolen) | **works** | ✓ |

## Test plan

- [x] S&P, call window correct, exchange `PIN/HIL` → two county QSOs logged.
- [x] S&P, call window `NY4A` (wrong), exchange `PIN/HIL NY4I` → callsign updated to NY4I, two county QSOs logged.
- [x] Normal single-county callsign correction (`599 PIN K1ABC`) unchanged.
- [x] FullBuild green (119 unit tests pass).

## Out of scope / follow-up

No automated test coverage was added — `LooksLikeACallSign` lives in `trdos/LOGSTUFF.PAS` which the test runner does not link.  A future refactor that moves callsign-shape helpers into a dependency-light unit (along the lines of the Issue #887 `uADIF` extraction) would unblock automated testing here.

### DIFF
--- a/tr4w/src/trdos/LOGSTUFF.PAS
+++ b/tr4w/src/trdos/LOGSTUFF.PAS
@@ -2364,19 +2364,22 @@ begin
     end
     else if Call[CharIndex] = '/' then
     begin
-      if ((CharIndex > 3) and (CharIndex <> length(Call)) or
-        ((CharIndex < length(Call)) and (CharIndex > 1))) then
+      // A '/' at the very start or very end of the token cannot belong
+      // to a callsign (no prefix or no body to suffix).
+      if (CharIndex = 1) or (CharIndex = length(Call)) then
       begin
-        LooksLikeACallSign := True;
+        LooksLikeACallSign := False;
         Exit;
       end;
-    end;
-
-    if ((CharIndex = 1) or (CharIndex = length(Call))) and (Call[CharIndex] =
-      '/') then
-    begin
-      LooksLikeACallSign := False;
-      Exit;
+      // Otherwise: do NOT return True here just because we found a '/'.
+      // The legacy code returned True at the first interior '/' which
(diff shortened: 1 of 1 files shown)


n4af:
https://tr4w.net/4.147/tr4w_setup_4_147.03.exe

PR #896, opened 2026-05-11
Issue #887: extract ADIF parser/emitter into testable string-in/string-out functions

ny4i:
## Summary

Refactor TR4W's ADIF code into a focused, dependency-light unit (`uADIF.pas`)
with pure string-in / string-out entry points, so the ADIF format logic
can be exercised by unit tests independent of `MainUnit` and the rest
of the application.

The bulk of the moved logic (field-list lexer, field-name → enum mapping,
`ApplyADIFFieldsToExchange`, `EmitADIFRecord`, `ExportADIFToString`)
now lives in `tr4w/src/uADIF.pas`.  `MainUnit.ImportFromADIF` and
`PostUnit.ExportToADIF` are thin drivers around the pure functions,
with contest-specific tail emission (fields sourced from `MainUnit`
globals) supplied via a callback.

### What the PR contains

Eleven commits, landing in three logical groups:

**ADIF refactor (Issue #887, parts 1-4)**

- `dd471f2` part 1: extract ADIF field-list lexer + unit tests
- `d8a9e81` part 2: move `TADIF_Fields` enum and helpers from MainUnit
- `417eada` part 3: field-name mapping + `ImportADIFFromString`
- `37daddb`        import bug fixes uncovered by FQP round-trip testing
- `477d340` part 4a: export side + differential round-trip tests
- `c34c46a`        fixture-based import regression tests + two bug fixes
- `a677209` part 4b: refactor `PostUnit.ExportToADIF` to use `uADIF`

**Latent-bug fix (separate from the refactor; surfaced by it)**

- `ea6d33a` fix stack-buffer overflow in `sWriteFileFromString`

**Phase 1 ADIF export verifier (new test infrastructure)**

- `ebc4cfe` add ADIF export verifier (logdump + Python cross-checker)
- `c86950f` add CONTEST_ID, FREQ, APP_TR4W_ROVERCALL field checks
- `5e07651` logdump: detect non-aligned .TRW files up front

### Real latent TR4W bugs found and fixed by this work

Three pre-existing bugs that were hiding in the codebase, surfaced by
the refactor + testing work:

1. **`sWriteFileFromString` stack-buffer overflow** (commit `ea6d33a`,
   `tr4w/src/utils/utils_file.pas`).  The function copied its input
   into a fixed 256-byte stack buffer via `StrLCopy(buffer, ..., 255)`,
   then asked `WriteFile` to write `Length(sBuffer)` bytes from that
[...]

PR #899, opened 2026-05-12
K4 PING/PONG keep-alive + ADIF SRX_STRING / SRX-STX fixes

### LINKED ISSUE #897: Implement PING/PONG KeepAlive for network K4
Per the programmer's reference manual:

PING & PONG (Connection Test / Keep-Alive)

        

SET format: PING;

        RESP format: PONG;

"PING;" can be used to check connection status. It can also be used to keep a link alive. The server will drop clients that are unresponsive for 10 seconds. Clients should send the “PING;” command once per second to keep data flowing.

### LINKED ISSUE #898: ADIF SRX_STRING wrongly prepends RST for contests where the exchange has no RST
## Summary

`<SRX_STRING>` in the ADIF export incorrectly prepends `RSTReceived` to the exchange string for contests whose exchange format does not include an RST. The bug is visible in any contest where `ExchangeInformation.RST = False` — ARRL Field Day, Winter Field Day, ARRL Sweepstakes (CW/SSB), etc.

## Reproduction

Export a small ARRL Field Day log (e.g. one QSO `<CALL>K6XX <CLASS>1D <ARRL_SECT>WCF`) via File → Export to ADIF.

## Observed

```
<SRX_STRING:9>59 1D WCF
```

## Expected

```
<SRX_STRING:6>1D WCF
```

(The `59` is the default `RSTReceived` value; Field Day's exchange is class + section only.)

## Cause

`uADIF.ResolveSRXString` was added during the Issue #887 refactor (commit f048dc7) to normalize `<SRX_STRING>` to include the implied RST for state QSO parties (so it's symmetric with `<STX_STRING>` regardless of whether the operator typed `599`). That assumption is wrong for contests whose exchange format does not include RST.

The contest's exchange shape is described by the existing per-contest `ExchangeInformation.RST: Boolean` in `trdos/LOGDUPE.PAS` (set in `SetUpExchangeInformation`). uADIF can't read it directly without taking on a trdos dependency, which would defeat the testability premise of #887.

## Proposed fix

Move `<SRX_STRING>` emission out of `uADIF.EmitADIFRecord` and into `PostUnit.EmitContestSpecificTailForExport` — the established tail-emitter callback that already handles contest-specific ADIF fields (CNTY, CQZ, MY_POTA_REF, ARRL_SECT, CLASS, etc.). The tail emitter has direct access to `ExchangeInformation` and can pick the right shape:

- POTA: park-ref shape (existing branch, unchanged)
- `ExchangeInformation.RST = True`: normalize to `"RST + exch"`  (state QPs / zone contests — current behaviour for those)
- `ExchangeInformation.RST = False`: emit `ExchString` as-is  (FD, SS, etc.)

## Out of scope

Other contest-specific ADIF fields are not affected — this is specifically about the SRX_STRING shape decision.

ny4i:
## Summary

Three independent fixes that surfaced while doing Field Day prep testing:

1. **K4 PING/PONG keep-alive in network mode** (closes #897)
2. **ADIF `<SRX_STRING>` no longer prepends `59` for FD / SS / Winter FD** (closes #898)
3. **ADIF `<SRX>` / `<STX>` no longer emit garbage like `<SRX:5>000-1` for unset serials** (no issue — surfaced during the FD export verification for #898)

## 1. K4 PING/PONG keep-alive (`uRadioElecraftK4.pas`, Issue #897)

A K4 server (K4 in remote/host mode, `k4remote.elecraft.com`, etc.) drops a client that sends nothing for 10 seconds. TR4W's K4 network mode uses AI5 — state changes are pushed from the radio to the client — but the client itself sends nothing during operator idle periods. An operator who tunes once and then sits quietly for >10 s would silently lose the K4 connection.

Fix: in network mode, enable the existing polling infrastructure at 1 Hz and use it to send `PING;`. Serial mode is unchanged (still sends `IF;FB;` for state polling, since AI is disabled on serial to avoid flooding the COM port).

- `Connect` (network branch): `requiresPolling := True; pollingInterval := 1000`.
- `PollRadioState`: branches on `serialPort` — serial sends `IF;FB;` (current), network sends `PING;`.
- `ProcessMessage`: `'PO'` added to `AnsiIndexText` list with an explicit case 17 that logs the PONG keep-alive response. PONG also refreshes the inbound watchdog via the existing `UpdateLastValidResponse` at the top of `ProcessMessage`. Explicit case branch (rather than fall-through) so PONG appears in the dispatch list alongside every other K4 message — a future reader sees it's handled, not silently dropped.

## 2. `<SRX_STRING>` no longer prepends `59` for FD / SS / Winter FD (`uADIF.pas` + `PostUnit.PAS`, Issue #898)

`ResolveSRXString` was added in PR #896 to normalize `<SRX_STRING>` with an implied RST prefix so it would be symmetric with `<STX_STRING>` for state QSO parties (where the operator types just the county and the RST is implied 599). That assumption is wrong for any contest whose exchange does not include RST: ARRL Field Day, Winter Field Day, ARRL Sweepstakes (CW/SSB), etc. For Field Day, the exchange is class + section (e.g. `1D WCF`) and the emit was producing `<SRX_STRING:9>59 1D WCF` — the leading `59` is bogus.

The "does this contest's exchange include RST?" knowledge already lives in `ExchangeInformation.RST` (`trdos/LOGDUPE.PAS`), which is set per-contest by `SetUpExchangeInformation`. `uADIF` cannot read that without taking on a trdos dependency, which would defeat the testability premise of Issue #887.

Fix: move `<SRX_STRING>` emission out of `uADIF.EmitADIFRecord` and into `PostUnit.EmitContestSpecificTailForExport` — the established tail-emitter callback for contest-specific ADIF fields, which already handles POTA's park-ref-shaped `<SRX_STRING>`. The tail emitter now decides the `<SRX_STRING>` shape:

| Case | Behaviour |
|---|---|
| `POTA` | park-ref shape (existing branch, unchanged) |
| `ExchangeInformation.RST = True`  | `ResolveSRXString` — `"RST + ExchString"` (state QPs / zone contests) |
| `ExchangeInformation.RST = False` | `ExchString` as-is (FD, SS, Winter FD, ...) |

`ResolveSRXString` is promoted from a private helper to a public function in `uADIF`'s interface so the tail emitter can call it.

## 3. `<SRX>` / `<STX>` guard against unset-serial sentinels (`uADIF.pas`)

Two distinct "unset" sentinels for `NumberReceived` / `NumberSent` are in use:

| Sentinel | Source |
|---|---|
| `$FFFF` (65535) | `uADIF.InitContestExchangeForParse` for records built from imported ADIF |
| `-1`            | live-entry / binary-log paths for contests that don't use a serial |

The prior guard was just `<> $FFFF`, written when these fields were a `Word` type (where `$FFFF` was the natural -1-as-unsigned representation). The fields are now signed `Integer`, so `-1` stays `-1` — the `$FFFF` check no longer catches it. For Field Day, where SRX is unused and stays at `-1`, `EmitADIFRecord` was producing `<SRX:5>000-1`: `PadInt(-1, 5)` calls `IntToStr(-1) = '-1'` and then zero-pads to width 5, yielding the literal `"000-1"`. That's malformed ADIF for a numeric field.

Fix: require the value to be positive AND not the parse sentinel. Same guard applied symmetrically to STX so future code paths that leave `NumberSent` unset don't produce the same garbage.

## Test plan

- [x] Full build clean (`FullBuild.ps1`)
- [x] All 561 unit tests pass (`tr4w_unit_tests.exe`)
- [x] Live Field Day export — verified: `<SRX_STRING:6>1D WCF` instead of `<SRX_STRING:9>59 1D WCF`; no `<SRX:5>000-1` garbage; `<STX:5>00001` still emitted for the operator's auto-incremented serial; second `<STX_STRING>` (legacy `MyFDClass + MySection` form) still emitted as before.
- [x] State QP export regression-check — `<SRX_STRING>` still has the RST prefix as before; SRX with positive serials still emitted.
- [x] **K4 network keep-alive — verified on real K4 hardware.** Connection stayed up through idle periods >10 s. `tr4w.log` at TRACE/DEBUG showed `[ProcessMessage] PONG keep-alive response` ~1× per second as expected.

### DIFF
--- a/tr4w/src/trdos/PostUnit.PAS
+++ b/tr4w/src/trdos/PostUnit.PAS
@@ -1787,6 +1787,26 @@ begin
    Result := Result + EmitADIFField('STX_STRING',
                                     Trim(DeleteRepeatedSpaces(stxString)));
 
+   // ----- SRX_STRING (Issue #898) -----
+   // uADIF.EmitADIFRecord deliberately does NOT emit SRX_STRING because
+   // its shape is contest-specific:
+   //   POTA          -> park-ref shape (handled below in the POTA branch)
+   //   RST-implied   -> "RST + ExchString"  (ResolveSRXString normalizes)
+   //   no RST        -> ExchString as-is    (FD, SS, Winter FD, etc.)
+   // The "is RST part of this contest's exchange?" knowledge lives in
+   // ExchangeInformation.RST (trdos/LOGDUPE.PAS), which uADIF cannot
+   // see without taking a trdos dependency.  POTA's own SRX_STRING is
+   // emitted further down -- this branch handles every other contest.
+   if rec.ceContest <> POTA then
+      begin
+      if ExchangeInformation.RST then
+         Result := Result + EmitADIFField('SRX_STRING',
+                                          ResolveSRXString(rec))
+      else
+         Result := Result + EmitADIFField('SRX_STRING',
+                                          string(rec.ExchString));
+      end;
+
    // ----- POTA-specific (uADIF skips SRX_STRING for POTA) -----
    if rec.ceContest = POTA then
       begin
--- a/tr4w/src/uADIF.pas
+++ b/tr4w/src/uADIF.pas
@@ -1321,10 +1321,21 @@ begin
          Result := Result + EmitADIFField('RX_PWR', string(rec.Power));
       end;
 
-   // SRX / STX (numeric serials)
-   if rec.NumberReceived <> $FFFF then       // -1 as Word -> not set
+   // SRX / STX (numeric serials).  Two "unset" sentinels are in use:
+   //   $FFFF (65535) -- set by uADIF.InitContestExchangeForParse for
+   //                    records built from imported ADIF
+   //   -1            -- left in place by live-entry / binary-log
+   //                    paths for contests that don't use a serial
+   //                    (e.g. Field Day on the SRX side -- ExchString
+   //                    carries class+section, NumberReceived stays
+   //                    at its initialized value)
+   // Without the >0 guard, FD records emit garbage like <SRX:5>000-1
+   // because PadInt(-1, 5) produces the literal string "000-1".
+   if (rec.NumberReceived > 0)      and
+      (rec.NumberReceived <> $FFFF) then
       Result := Result + EmitADIFField('SRX', PadInt(rec.NumberReceived, 5));
-   if rec.NumberSent <> $FFFF then
+   if (rec.NumberSent > 0)      and
+      (rec.NumberSent <> $FFFF) then
       Result := Result + EmitADIFField('STX', PadInt(rec.NumberSent, 5));
 
    if rec.TenTenNum <> $FFFF then
--- a/tr4w/src/uADIF.pas
+++ b/tr4w/src/uADIF.pas
@@ -257,6 +257,16 @@ function ExportADIFToString(const records: TContestExchangeArray;
 // reachable from both uADIF (export) and the test runner.
 function GetStateForContest(c: ContestType): string;
 
+// Build the RST-normalized SRX_STRING value: if ExchString already
+// starts with the literal RSTReceived (e.g. operator typed `599 HIL`),
+// return ExchString unchanged; otherwise prepend `RSTReceived ` so
+// the field is symmetric with STX_STRING.  Used by the export tail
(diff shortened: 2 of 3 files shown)


n4af:
https://tr4w.net/4.147/tr4w_setup_4_147.05.exe should include new
tr4wserver.exe

Issue #900, opened 2026-05-12
Double-click column divider auto-fit truncates content on next restart


ny4i:
## Summary

When the operator double-clicks a column divider in the editable log view to auto-fit, the column visually fits its content correctly that session.  After saving (via the existing `HDN_ENDTRACK` save path) and restarting TR4W, the same column reapplies the saved width but the cell content now displays truncated with an ellipsis -- even though it fit pre-restart at "the same" pixel width.

## Reproduction

1. Resize the editable log columns however the operator likes.
2. Double-click the right border of (e.g.) the Freq column header.  The column shrinks to just-fit `14046.00` etc.  Width displays correctly.
3. Close TR4W.
4. Restart.
5. Observe that `14046.00` now displays as `14046...` -- truncated.

## Root cause

The Win32 ListView/Header control's default `HDN_DIVIDERDBLCLICK` auto-fit picks a width based on the header-divider position.  That value is a few pixels narrower than the cell-content area because the ListView's cell layout adds an internal left/right margin.  At the just-fits threshold, the saved value is enough to display the content visually in the current session (the column was rendered before save) but applying the same pixel value on restart triggers ellipsis truncation.

A second wrinkle uncovered during testing: `HDN_ENDTRACK` only fires for end-of-drag, not for the OS's internal auto-fit during `HDN_DIVIDERDBLCLICK`.  So a "set a flag in `HDN_DIVIDERDBLCLICK`, add padding in `HDN_ENDTRACK`" approach silently drops the save on double-click entirely -- the previous CFG value persists across restart.

## Fix

Intercept `HDN_DIVIDERDBLCLICK` directly in `tr4w.dpr`'s `WindowProc`:

1. Run the auto-fit ourselves via `LVSCW_AUTOSIZE_USEHEADER`.
2. Read the resulting width.
3. Add a small padding constant (12 pixels) for breathing room.
4. Apply the padded width to the column.
5. Save the padded width via `SaveColumnWidthToConfig`.
6. Return non-zero `Result` and `Exit` to suppress the OS's default auto-fit (which would otherwise overwrite the padded width).

Manual drag remains unchanged -- `HDN_ENDTRACK` saves the dragged width exactly as the operator chose it.  Padding is applied only to double-click auto-fit.

## Affects

- `tr4w/tr4w.dpr` -- `WindowProc` WM_NOTIFY header handler

## Out of scope

[...]

Issue #750, opened 2024-04-02
Add ability to mark a QSO as an X-QSO


ny4i:
Sometimes we do not want to delete a QSO but do not want it to be counted a QSO in the Cabrillo file. Cabrillo allows for the QSO to be marked X-QSO in this case. This could be a case where one works a station incorrectly (extra band change or working a DX station in the middle of the contest--non-contest related station that is).

QSOs can be marked deleted in the Edit dialog, but there should be another option to mark it as an X-QSO. This should go in the ContestExchange (use a reserved field).

When the Cabrillo file is created, if x-qso is true, emit X-QSO instead of QSO in that entry. The QSO should be exported from ADIF as normal.

Also, when the QSO is marked an X-QSO, the LogContactToUDP should set IsClaimedQSO to 0 instead of 1.

PR #901, opened 2026-05-13
May 2026 enhancements: X-QSO + column auto-fit padding + WSJT-X mult-check gate

ny4i:
## Summary

Three commits, three independent concerns, all tested live.

1. **Issue #750** — Mark a QSO as Cabrillo X-QSO (logged but not claimed for credit)
2. **Issue #900** — Pad column auto-fit on double-click so content survives restart
3. **WSJT-X cleanup** — Skip mult checks for non-multiplier contests (no issue, cosmetic)

## 1. Issue #750 — X-QSO support (`85800fe`)

X-QSO records stay in the log for NIL protection of the worked station but contribute nothing to the score -- no QSO count, no multipliers, no points, no presence in the dupe sheet.  The Cabrillo export emits an `X-QSO:` line prefix instead of `QSO:`, ADIF export marks the record via `APP_TR4W_CLAIMEDQSO=0`, and the editable log paints the row grayed so the operator can see at a glance which contacts are excluded.

| Aspect | Decision |
|---|---|
| Storage | `ceXQSO: Boolean` in `ContestExchange`, freed from `sReserved` (record size unchanged; legacy logs read as `ceXQSO=False`) |
| Edit dialog | New checkbox after S&P |
| Visual indicator | Full row grayed out in editable log view (mid-gray text via `NM_CUSTOMDRAW`) |
| Dupes / mults / points | Not counted |
| ADIF emit | `APP_TR4W_CLAIMEDQSO` (`1` = normal, `0` = X-QSO), TR4W field only |
| ADIF import | Recognizes `APP_TR4W_CLAIMEDQSO`, `APP_N1MM_CLAIMEDQSO`, `APP_DXLOG_XQSO` for cross-tool round-trip |
| Cabrillo | Leading tag becomes `X-QSO:` instead of `QSO:` |
| UDP | `LogContactToUDP` reports `<IsClaimedQso>0</IsClaimedQso>` for X-QSO records |
| Right-click | Deferred (no existing right-click infra in the editable log) |

**Score / dupe / mult exclusion is enforced at both rebuild paths:** `tUpdateLog(actRescore)` and `LoadinLog`.  X-QSO records also have `QSOPoints` zeroed at rescore (DXLog.net convention -- consistent visual `0` in the Pts column regardless of what the contact was worth before being marked).  Toggle X-QSO off and a normal rescore restores the points based on contest rules.

Tests: 632 unit tests pass (added `Test_RoundTrip_XQSO_Field`, `Test_Import_XQSO_N1MM_Convention`, `Test_Import_XQSO_DXLog_Convention` plus a new fixture covering all three field conventions).

**Resource files:** the X-QSO checkbox (resource ID 170) is now in `tr4w_eng.RES` plus all 10 other language `.RES` files (rus, ser, esp, mng, pol, cze, rom, chn, ger, ukr).  Pascal-side wiring (`FLD_XQSO = 170` and Set/GetCheck handlers in `uEditQSO.pas`) activates the checkbox automatically.

## 2. Issue #900 — Column auto-fit padding (`a795c86`)

When the operator double-clicks a column divider in the editable log to auto-fit, the Win32 Header control picks a width based on the divider position -- a few pixels narrower than the cell-content area because the ListView's cell layout adds an internal margin.  At the just-fits threshold, the value displays content correctly in the current session but applying the same pixel value on restart triggers ellipsis truncation (e.g. `14046.00` -> `14046...`).

A second issue uncovered during testing: the OS's internal auto-fit during `HDN_DIVIDERDBLCLICK` does **not** generate an `HDN_ENDTRACK`, so a flag-based \"set in DBLCLICK, save in ENDTRACK\" approach silently drops the save on double-click.

Fix: intercept `HDN_DIVIDERDBLCLICK` directly in `WindowProc`:

1. Run auto-fit via `LVSCW_AUTOSIZE_USEHEADER`.
2. Read resulting width.
3. Add `COLUMN_DOUBLECLICK_PAD_PX` (12 px).
4. Apply the padded width.
5. `SaveColumnWidthToConfig`.
6. Return `Result := 1` and `Exit;` to suppress the OS's default auto-fit.

Manual drag is unchanged -- `HDN_ENDTRACK` still saves the dragged width exactly.

## 3. WSJT-X mult-check gate (`c1c369e`)

The CQ-decode handler in `uWSJTX.pas` was calling `DetermineIfNewMult` and `DetermineIfNewDomesticMult` on every decode regardless of whether the current contest had any multiplier dimension active.  For ARRL-DIGI / CQ-WPX-DIGI / similar QSO-count-only contests, every `Doing*Mults` global is False so these lookups always return False -- but they still produced confusing log lines like `[uWSJTX] Checking if grid FN is a multiplier` and `is NOT a domestic MULT` at DEBUG level.

Gate the multiplier branches behind `DoingDomesticMults or DoingDXMults or DoingPrefixMults or DoingZoneMults`.  Dupe handling is unchanged.  Operator-visible behaviour matches the original design: highlight only the exceptions (dupes to avoid, mults to prioritize); work everyone else by default with no highlight.

No functional change for contests with mults; observable change is the absence of misleading debug log lines for no-mult contests.

## Test plan

- [x] Full build clean (`FullBuild.ps1`)
- [x] All 632 unit tests pass
- [x] X-QSO end-to-end:
  - [x] Edit dialog checkbox toggles on/off, persists in log
  - [x] Visual gray-out applies on row when X-QSO; reverts on untoggle
  - [x] `QSOPoints` shows 0 for X-QSO records, restored when untoggled
  - [x] Multi-record QSO totals correctly exclude X-QSO records (live + after restart)
  - [x] Cabrillo export emits `X-QSO:` prefix for X-QSO records, `QSO:` for normal
  - [x] ADIF export emits `<APP_TR4W_CLAIMEDQSO:1>0` only for X-QSO records
  - [x] ADIF round-trip preserves flag (TR4W -> TR4W)
  - [x] UDP `<IsClaimedQso>` is `0` for X-QSO and skipped QSOs, `1` otherwise
- [x] Column auto-fit padding:
  - [x] Double-click divider produces visibly wider column than pre-fix (no ellipsis)
  - [x] Saved width persists across restart with content fully visible
  - [x] Manual drag still saves the dragged width exactly (no padding added)
- [x] WSJT-X cleanup:
  - [x] ARRL-DIGI log: misleading mult-check log lines no longer fire on CQ decode
  - [x] Dupe highlighting still works in contests both with and without mults
- [x] X-QSO checkbox (ID 170) added to all 10 remaining language `.RES` files (`tr4w_rus`, `tr4w_ser`, `tr4w_esp`, `tr4w_mng`, `tr4w_pol`, `tr4w_cze`, `tr4w_rom`, `tr4w_chn`, `tr4w_ger`, `tr4w_ukr`).


n4af:
https://tr4w.net/4.147/tr4w_setup_4_147.06.exe

Issue #904, opened 2026-05-14
Migrate away from ICOM NETWORK USER and PASSWORD


ny4i:
In preparation for other network radios that require a user and/or password, I fear that our choice of the command ICOM NETWORK USERNAME and ICOM NETWORK PASSWORD are too specific. If we dropped the ICOM prefix, that would make it  more generic. Of course, the ICOM fields are already in the wild, so we have to accept those too. An example can be found in the Kenwood TS890 implementation

n4af:
https://tr4w.net/4.147/tr4w_setup_4_147.08.exe

Issue #436, opened 2020-08-03
Add rig control to TS890


ny4i:
This radio resembles the TS990 more than any other so reference the issues with #435.

ny4i:
https://www.kenwood.com/i/products/info/amateur/pdf/ts890_pc_command_en_rev1.pdf

ny4i:
This radio also supports a network connection. See this doc for more info and implementation details: https://github.com/ny4i/TR4QT/blob/master/docs/kenwood-direct-connection-flow.md

ny4i:
Testing report from @n2skh

Tom, I finally had some time today to test TR4W with the 890.  Since I'd been using the USB connection with the Kenwood remote software (ARCP-890), I first reconfigured it to use TCP/IP and confirmed the radio CAT interface was working, then I installed and configured TR4W.  When TR4W started the radio appeared to connect, but my 890 immediately gave me a message on the screen that the bandscope was out of range. It then blanked the screen after a few seconds and put up a message to turn off the DC power, which I did.  When I turned it back on, the bandscope error was still there (and I noticed the bandscope range was from 0 to 0).  I was able to clear the message and restore the bandscope by hitting the scope range button. (edited) 
[11:15 PM]I tried starting TR4W again, but this time I turned off the bandscope on the radio first and I didn't get that same message, but the screen would blank and reset every 5 to 10 seconds (I didn't time it), and I noticed the radio 1 indicator on the main TR4W screen would go from black to red and then back to black when this happened, so I guessed the connection was being lost and reconnected.  I confirmed this by looking at the log. (edited) 
[11:16 PM]I never did get the frequency, etc from the radio to display on TR4W, so even though it's connecting, it's not quite right yet.
[11:18 PM]Here's the log file for one connection cycle (without the bandscope turned on).

[tr4w.log](https://github.com/user-attachments/files/27971926/tr4w.log)

ny4i:
## Field test report — N2SKH, 2026-05-28 (v4.147.15-era build)

Ron N2SKH tested LAN control against his TS-890S at 192.168.50.195:60000 with `DEBUG LOG LEVEL = TRACE`. Reconnect loop is stable but the session never reaches CAT.

### Observed pattern (repeats every cycle)

1. TCP connect → OK
2. TX `##CN;` → RX `##CN1;` (radio requests credentials)
3. TX `##ID00507N2SKH<pw>;` → RX `##ID1;` (credentials accepted)
4. **Total silence from the radio for ~21 s**
[...]

Issue #435, opened 2020-08-03
Implement Kenwood TS990 (and TS890 when added) Mode properly


ny4i:
Kenwood changed commands for the TS890 and TS990. Instead of MD; , they have an OM command. The OM command has Digital better enumerated so fix this for WSJT-X (in case we have any 990 or 890 users).

Split is done with the TB command.

Also in a massive step backwards, the 990 can not directly set the "Other" VFO **Mode** (according to Hamlib). We have to do the Icom switch-a-roo to bring the other to the forefront, set it and then flip it back. Quite unbelievable frankly.

> From Hamlib Kenwood.c comment:
> 
> /* The TS990s has targetable read mode but can only **set the mode**
>            of the current VFO :( So we need to toggle the operating VFO
>            to set the "back" VFO mode. This is done here rather than not
>            setting caps.targetable_vfo to not include
>            RIG_TARGETABLE_MODE since the toggle is not required for
>            reading the mode. */

The 990 and 890 do not appear to have an IF command so this is done with the FA; and FB; commands.

I would recommend making a new function call so instead of pKenwood2, use Kenwood_TS890_TS990 and use that to customize this stuff. We have too much in the Kenwood function anyway.

ny4i:
https://www.kenwood.com/i/products/info/amateur/pdf/ts_990_pc_command_e.pdf

ny4i:
This has been implemented and tested

PR #905, opened 2026-05-14
Generic network credentials + Kenwood TS-890 native control (Issues #904, #436)

ny4i:
## Summary

Two issues bundled because the second drives the first:

- **#904** — Drop the `ICOM` prefix from the network-credential config statements; `NETWORK USERNAME` / `NETWORK PASSWORD` is the new canonical name with the old `ICOM NETWORK ...` names retained as backward-compat aliases so existing `tr4w.ini` and `.cfg` files keep working.
- **#436** — Native TR4W control for the Kenwood TS-890S, both serial CAT (treated like a TS-990 in every Kenwood case statement) and direct TCP/IP over the radio's LAN port using the documented `##CN;` / `##ID0…;` authentication handshake. The TS-890 was previously HamLib-only.

Plus a small dropdown-ordering fix surfaced during testing.

## Commits

| Hash | Scope |
|---|---|
| `dc33056` | Issue #904 — Generic NETWORK USERNAME/PASSWORD with backward-compat aliases |
| `abb7e21` | Issue #436 — TS-890 native serial + new `TKenwoodTS890Radio` network class with auth handshake |
| `05095f0` | Fill the deferred TS-890 stubs: `SetRITFreq` / `SetXITFreq` (RC + RU/RDnnnnn), `SetFilter` (FL00/01/02) |
| `922f919` | Move TS890 / IC905 / IC7300MK2 to alphabetical positions in `InterfacedRadioType` so they show in the right place in the Radio Type dropdown |

Master also got a separate prereq fix (`fe307a4`, already merged) for the unterminated string literal in `Version.pas` from the 4.147.07 version-bump commit — `FullBuild.ps1` was failing for everyone before that landed.

## Issue #904 — Generic network credentials

**Why:** the `ICOM` prefix was leaky abstraction. The Kenwood TS-890 also needs LAN credentials and there is no good reason every future credentialed radio should carry "Icom" in its config statement name.

**What changed:**
- `Radio*.IcomNetworkUsername` / `IcomNetworkPassword` storage fields renamed to `Radio*.NetworkUsername` / `NetworkPassword` (LOGRADIO.PAS)
- New canonical config statements `RADIO ONE/TWO NETWORK USERNAME` and `RADIO ONE/TWO NETWORK PASSWORD` (uCFG.pas)
- Old `RADIO ONE/TWO ICOM NETWORK USERNAME/PASSWORD` statements kept as aliases at the same storage addresses — existing user `.cfg` / `tr4w.ini` files keep parsing correctly
- Dialog labels in the CAT config window changed from "ICOM USERNAME" / "ICOM PASSWORD" to "NETWORK USERNAME" / "NETWORK PASSWORD"
- On Save (`RestartPollingThread` in uCAT.pas) writes the canonical generic name to `tr4w.ini` AND deletes the legacy `ICOM NETWORK …` key — so once the user saves once, the .ini is clean

### Native serial (LOGRADIO.PAS)

The TS-890 was `rt:rtHamlib` in `RadioParametersArray`. Changed to `rt:rtKenwood` and added to every Kenwood case statement that previously listed TS990, plus the `KenwoodRadios`, `RadioSupportsCWByCAT`, `RadioSupportsCWSpeedSync`, and `RadioSupportsPlayDVK` sets. Dropped from `HamLibONLYRadios`. Default baud 4800 (radio menu can raise to 115200).

### Native network (new `src/uRadioKenwoodTS890.pas`)

`TKenwoodTS890Radio` extends `TNetRadioBase`, modeled on `TK4Radio`. Speaks the standard Kenwood ASCII CAT (semicolon-terminated FA/FB/OM/KS/RC/RU/RD/etc.) over a TCP socket after a 3-step LAN auth handshake:

```
##CN;                              -->  ##CN1;
##ID0<idLen:2><pwLen:2><id><pw>;   -->  ##ID1;
```

- Default port: **60000**
- Auth state machine: `ksNone` → `ksWaitingForCN` → `ksWaitingForID` → `ksAuthenticated` (or `ksAuthFailed`). All CAT operations are no-ops until authenticated. Empty `NetworkUsername` skips the handshake (future simulator path).
- Post-auth init: `AI2; FA; FB; OM0; OM1; KS; TB; FT; RT; XT; ID;` — `ID;` must return `ID024;` for TS-890S
- AI2 push handles freq/mode updates; `requiresPolling := False`
- TS-890 OM mode map per Kenwood PC Command Reference Rev 1: `1=LSB 2=USB 3=CW 4=FM 5=AM 6=FSK 7=CW-R 9=FSK-R A=PSK B=PSK-R C/D/E/F=LSB/USB/FM/AM-DATA`
- CW speed 4-60 wpm via `KSnnn;`
- RIT/XIT: `RC;` clears, `RU<nnnnn>;` / `RD<nnnnn>;` sets a specific offset (silently capped at 9999). RIT and XIT share one offset on this radio.
- Filter slots A/B/C via `FL00/01/02`
- Memory keyer PB1..PB6

### Factory wiring

- New `TRadioModel.rmKenwoodTS890` enum entry in `uRadioFactory.pas` with `CreateRadioNetwork` case, `ModelToString`, `IsModelSupported`, and `GetSupportedModels` updated
- `RadioObject.MapRadioModelToFactory` (LOGRADIO.PAS) routes `TS890` → `rmKenwoodTS890`
- `RadioObject.SetUpRadioInterface` sets `NetworkUsername` / `NetworkPassword` on the new class before `Connect`, same pattern as the Icom branch
- `uCAT.pas` dialog shows the credential fields when the operator picks TS890 + TCP/IP port (added TS890 to the existing Icom-network set)

## Dropdown alphabetization fix

TS890, IC905, IC7300MK2 were each appended at the end of `InterfacedRadioType` when they were added — operators could not find TS890 in the Radio Type combo because it sat below the HamLib-only block. Moved each to its alphabetical position in the enum and in the two parallel arrays (`InterfacedRadioTypeSA`, `RadioParametersArray`). Pascal `case` statements match enum members by name not ordinal, so no case statement needed changing. As a positive side effect, `[IC78..IC9700, OMNI6]` patterns in `uProcessCommand.pas` and `uRadioPolling.pas` — which had been silently skipping IC905 and IC7300MK2 — now correctly include them.

## Build

`FullBuild.ps1` clean (unit tests + main exe both compile). No new warnings.

## Test plan

Most of this needs hardware; the network class implementation is grounded in the TR4QT reference flow doc and the Kenwood PC Command Reference rather than a live capture, so the auth handshake should be smoke-tested before relying on it.

- [x] **Config-rename round-trip (no radio needed):** open the CAT dialog, enter values into the NETWORK USERNAME/PASSWORD fields, click OK, verify `tr4w.ini` has the new `RADIO ONE NETWORK USERNAME=…` key and the old `RADIO ONE ICOM NETWORK USERNAME` key is gone. Reload, verify the fields repopulate.
- [x] **Backward-compat alias (no radio):** manually edit `tr4w.ini` to add `RADIO ONE ICOM NETWORK USERNAME=…`, restart TR4W, verify the value is read and displayed. Save and verify the .ini now has the new name only.
- [x] **Dropdown alphabetization (no radio):** open the CAT dialog. TS890 should appear between TS870 and TS940. IC905 between IC781 and IC910. IC7300MK2 right after IC7300. No more orphans at the end past HAMLIB-ANY.
- [ ] **TS-890 serial (radio required):** select TS890, COM port, baud 4800 (or whatever the radio's menu is set to). Verify frequency / mode track the radio. RIT shift up/down works. CW speed sync works (if equipped).
- [ ] **TS-890 LAN auth (radio required):** in the radio's menu, set Admin ID and Password under LAN settings. In TR4W, set RADIO TYPE = TS890, port = TCP/IP, RADIO ONE IP ADDRESS = the radio's IP, RADIO ONE TCP PORT = 60000, fill NETWORK USERNAME/PASSWORD. Connect; verify the `tr4w.log` shows "LAN auth: sent ##CN;" → "received ##CN1, sent ##ID credentials" → "Authenticated successfully" → "Confirmed TS-890S (ID024)". If the password is wrong, verify it logs "AUTH FAILED" with the radio's response.
- [ ] **TS-890 LAN runtime (radio required):** turn the VFO; TR4W's frequency display should track via AI2 push. Set frequency from TR4W; radio tunes. Set RIT offset to a specific value; verify on radio display.

## Deferred / known limitations

- `TKenwoodTS890Radio.ToggleMode` / `ToggleBand` left as no-op stubs matching the K4's pattern — no native single-shot toggle commands and not exercised by TR4W's current keystroke flow
- `SetFilterHz` left as no-op — the TS-890 has bandwidth commands separately from the FL filter-slot selection; can be added in a follow-up if we want bandwidth-by-Hz control
- `SetMode(mode, nrVFOB)` doesn't actually set VFO B's mode because the TS-890 `OM` setting command ignores its P1 (VFO) byte per the PC Command Reference — documented in code as a known Kenwood quirk
- TS-990 still uses `MD;` instead of `OM;` — the TS-890 spec notes OM applies to TS-990 too, but switching the existing TS-990 native path is left as a separate cleanup (Issue #435)

## Wiki

To follow as a separate change: add TS-890 row to `Supported-Transceivers.md` with Network = "Kenwood Net", document the Admin ID/Password + port 60000 setup, add TS-890 to the CW-by-CAT page.

### DIFF
--- a/tr4w/src/VC.pas
+++ b/tr4w/src/VC.pas
@@ -979,6 +979,7 @@ type
     TS690,
     TS850,
     TS870,
+    TS890,    // Issue #436 -- alphabetical placement among Kenwoods
     TS940,
     TS950,
     TS990,
@@ -1037,12 +1038,14 @@ type
     IC765,
     IC775,
     IC781,
+    IC905,    // Alphabetical placement among Icoms (between IC781 and IC910)
     IC910,
     IC970D,
     IC7000,
     IC7100,
     IC7200,
     IC7300,
+    IC7300MK2,  // Alphabetical placement among Icoms (right after IC7300)
     IC7410,
     IC7600,
     IC7610,     // 4.65.7 issue 282
@@ -1058,11 +1061,8 @@ type
     FLRIG,
     TRXMANAGER,
     FT757GXII,
-    TS890,
     EXPERTTCI,
     ACLOG,
-    IC905,
-    IC7300MK2,
     HAMLIBANY);
 
 
--- a/tr4w/src/trdos/LOGRADIO.PAS
+++ b/tr4w/src/trdos/LOGRADIO.PAS
@@ -496,6 +496,7 @@ const
     ({Name: 'TS690';      } BR:BR4800;  P: 1; C: 1; S: 1; T:1; RA: $00; SW: 1; SFOC: $00; SMOC: $00; SPOC: $00; RX: $00; TX: $00; MB: 0; CW: $00;  LSB: $00; USB: $00; FM: $00; AM: $00; DIGL: $00; DIGU: $00; hamlibID: 2005; rt: rtKenwood),
     ({Name: 'TS850';      } BR:BR4800;  P: 1; C: 1; S: 1; T:1; RA: $00; SW: 1; SFOC: $00; SMOC: $00; SPOC: $00; RX: $00; TX: $00; MB: 0; CW: $00;  LSB: $00; USB: $00; FM: $00; AM: $00; DIGL: $00; DIGU: $00; hamlibID: 2009; rt: rtKenwood),
     ({Name: 'TS870';      } BR:BR4800;  P: 1; C: 1; S: 1; T:1; RA: $00; SW: 1; SFOC: $00; SMOC: $00; SPOC: $00; RX: $00; TX: $00; MB: 0; CW: $00;  LSB: $00; USB: $00; FM: $00; AM: $00; DIGL: $00; DIGU: $00; hamlibID: 2010; rt: rtKenwood),
+    ({Name: 'TS890';      } BR:BR4800;  P: 1; C: 1; S: 1; T:1; RA: $00; SW: 1; SFOC: $00; SMOC: $00; SPOC: $00; RX: $00; TX: $00; MB: 0; CW: $00;  LSB: $00; USB: $00; FM: $00; AM: $00; DIGL: $00; DIGU: $00; hamlibID: 2041; rt: rtKenwood),  // Issue #436 -- native Kenwood serial; default 4800 baud (radio menu can raise to 115200)
     ({Name: 'TS940';      } BR:BR4800;  P: 1; C: 1; S: 1; T:1; RA: $00; SW: 1; SFOC: $00; SMOC: $00; SPOC: $00; RX: $00; TX: $00; MB: 0; CW: $00;  LSB: $00; USB: $00; FM: $00; AM: $00; DIGL: $00; DIGU: $00; hamlibID: 2011; rt: rtKenwood),
     ({Name: 'TS950';      } BR:BR4800;  P: 1; C: 1; S: 1; T:1; RA: $00; SW: 1; SFOC: $00; SMOC: $00; SPOC: $00; RX: $00; TX: $00; MB: 0; CW: $00;  LSB: $00; USB: $00; FM: $00; AM: $00; DIGL: $00; DIGU: $00; hamlibID: 2012; rt: rtKenwood),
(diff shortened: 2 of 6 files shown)


ny4i:
Once built, I will ask a local friend to test it on his TS-890.

n4af:
https://tr4w.net/4.147/tr4w_setup_4_147.08.exe

PR #907, opened 2026-05-15
Fix IC-7760 disconnect color, recovery, and detection speed

ny4i:
## Summary

Fixes three compounding defects in Icom network disconnect handling that surfaced during IC-7760 testing. The user-visible symptom was: turn the radio off → ~15 s before the magenta "radio lost" indicator appeared → ~1 s later the indicator switched back to normal color (as if reconnected) → never actually recovered when the radio was turned back on.

Each defect alone is small. Together they made the radio appear connected when it was unreachable AND prevented automatic recovery when it returned.

### 1. Alert color flashed back to normal during reconnect

`TIcomRadio.GetIsConnected` returns `True` for any non-`Disconnected` transport state — that's deliberate so the polling thread doesn't interrupt the multi-step Icom handshake (`WaitingForHere` → `WaitingForReady` → `WaitingForLogin` → `Authenticated` → `StreamRequested` → `Connected`) by spamming `Connect()`. But `TIcomRadio` did **not** override `GetIsOperational`, so it inherited the base default of `True` unconditionally.

When the polling thread fired its 1-second-after-disconnect reconnect, the transport went `Disconnected → WaitingForHere`, the polling thread immediately took the "connected" branch, and `uRadioPolling.pas:781` unconditionally cleared the alert color (`SetRadioAlertState(False)`). Operator saw magenta flash off even though the radio was unreachable.

**Fix:** `TIcomRadio.GetIsOperational` now returns `True` only when `FNetworkTransport.State = icsConnected`. `uRadioPolling.pas:781` changed from `SetRadioAlertState(False)` to `SetRadioAlertState(not ro.IsOperational)`.

### 2. No recovery from a stuck handshake

The transport sends one `$0003` (AYH — "Are You Here") packet on `Connect()` and then sits in `WaitingForHere` forever if no `$0004` reply arrives. That's exactly what happens when `Connect()` fires while the radio is off. When the radio came back online, TR4W stayed silent because:

- The transport had given up on resending AYH (only sent once per `Connect()`)
- The polling thread's `IsConnected = True` (loose check, see #1) kept it in the connected branch instead of the else-branch that would have called `Connect()` again

**Fix:** new stuck-handshake detector at the top of `pNetworkRadio`'s connected branch. When `IsConnected = True && IsOperational = False` persists for >8000 ms, force a `Disconnect` so the next iteration's else-branch fires `Connect()` with a fresh AYH. Cycle repeats every ~9 s until the radio responds.

Gated on a new virtual `TNetRadioBase.GetCanRecycleOnStuckHandshake` that defaults `False`. `TIcomRadio` overrides to `True` for the network path. Flex (whose `IsOperational` tracks slice validity, not handshake state) and K4 / HamLib Direct (TCP-based, get disconnect detection from `socket.Connected`) all keep the default `False` so the detector is a no-op for them.

### 3. Disconnect detection took 15 seconds

`ICOM_PING_DEAD_TIMEOUT_MS` was `15000`. Radio sends pings every 100 ms, so 150 missed pings before declaring the link lost. Felt sluggish to the operator turning the radio off.

**Fix:** lowered to `3000`. 30 missed pings is unambiguous; transient sub-3-second network glitches still won't trigger a false reconnect cycle.

## Why Icom needs this and K4 doesn't

K4 is **TCP**. When you power off a K4 the OS sees the FIN/RST and `socket.Connected` becomes `False` essentially immediately — TCP carries connection state. The K4 polling path needs no application-level ping watchdog and gets disconnect detection for free.

Icom is **UDP** (`TIdUDPServer` instances for control and CI-V sockets per `uIcomNetworkTransport.pas:67-68`). UDP is connectionless — there's no socket-level disconnect event the OS can deliver. The only signal is the absence of the radio's expected ping packets. Hence the explicit timeout.

## Cross-radio impact (verified)

| Radio | Alert color | Stuck-recycle | Detection speed |
|---|---|---|---|
| **Icom (network)** | New: keyed off strict `IsOperational` | New: recycles after 8 s if mid-handshake | Was 15 s, now 3 s |
| Icom (serial) | No change (handshake is instant) | No-op (`CanRecycle = False` for serial) | N/A |
| K4 | No change (`IsOperational = True` from base → `not True = False`, identical to old `SetRadioAlertState(False)`) | No-op (`CanRecycle = False`) | TCP-driven, unchanged |
| FlexRadio | No change (already used `IsOperational` correctly via line ~944) | **Intentional no-op** — Flex's `IsOperational = FSlice0Valid` drops when SmartSDR closes; recycling TCP wouldn't fix that | TCP-driven, unchanged |
| HamLib Direct | No change | No-op | Unchanged |

## Files

| File | Change |
|---|---|
| `src/uNetRadioBase.pas` | New virtual `GetCanRecycleOnStuckHandshake` (default False); paired property |
| `src/uRadioIcomBase.pas` | `GetIsOperational` override (strict, true only when `State = icsConnected`); `GetCanRecycleOnStuckHandshake` override (true for network) |
| `src/uRadioPolling.pas` | Alert color line 781 keyed on `IsOperational`; new stuck-handshake detector at top of connected branch, gated on `CanRecycleOnStuckHandshake` |
| `src/uIcomNetworkTypes.pas` | `ICOM_PING_DEAD_TIMEOUT_MS` 15000 → 3000 |

## Test plan

Verified live on IC-7760:

- [x] **Disconnect detection time**: turning the radio off shows magenta in ~3 s (was ~15 s).
- [x] **Magenta stays magenta**: no false-positive flash back to normal color during reconnect attempts.
- [x] **Recovery on radio power-up**: with the radio off, log shows the polling thread recycling every ~9 s (`handshake stuck (IsConnected but not IsOperational) for >8000 ms; forcing Disconnect for retry`); when the radio is turned back on, the next reconnect attempt completes the handshake within ~7 ms and the alert clears.

Recommend smoke-testing on:
- [x] **K4 (network)** — confirm no behavior change. Power-off should still show magenta within 1–2 s (TCP-driven). Reconnect on power-up unchanged.
- [ ] **K4 (serial)** — confirm no behavior change.
- [ ] **FlexRadio (with SmartSDR open and active)** — confirm the in-progress fix doesn't trigger spurious recycles. Closing SmartSDR while TR4W is connected should still alert without forcing TCP reconnects.
- [ ] **HamLib Direct (any radio)** — confirm the polling loop is unaffected.

## Known limitations / not in scope

- The stuck-handshake recycle uses a fixed 8 s timeout and 1 s reconnect delay (~9 s cycle). Could be made configurable, but fixed values seemed reasonable for a first pass.
- The 3 s ping timeout is hard-coded in `uIcomNetworkTypes.pas`. If operators on flaky WiFi report false disconnects, we could promote it to a config statement.
- No new config statements added for this fix — pure behavior change.

### DIFF
--- a/tr4w/src/uIcomNetworkTypes.pas
+++ b/tr4w/src/uIcomNetworkTypes.pas
@@ -53,7 +53,7 @@ const
   ICOM_RETRANSMIT_CHECK_INTERVAL = 100;
   ICOM_CIV_WATCHDOG_INTERVAL     = 500;
   ICOM_CIV_TIMEOUT_THRESHOLD     = 2000;    // 2 seconds
-  ICOM_PING_DEAD_TIMEOUT_MS      = 15000;   // 15 seconds without a ping → full disconnect+reconnect
+  ICOM_PING_DEAD_TIMEOUT_MS      = 3000;    // 3 seconds without a ping -> declare link lost.  Radio sends pings every 100 ms, so 30 missed pings is unambiguous; a transient sub-3-second network glitch won't trigger a false reconnect cycle.  Was 15000 -- felt sluggish to operators turning the radio off (Issue: IC-7760 disconnect detection delay).
 
   // "Are You There" retry config
   ICOM_AYT_MAX_RETRIES    = 10;
--- a/tr4w/src/uNetRadioBase.pas
+++ b/tr4w/src/uNetRadioBase.pas
@@ -152,6 +152,13 @@ Type TNetRadioBase = class(TObject)
 
       function GetISConnected: boolean; virtual;
       function GetIsOperational: boolean; virtual;
+      // True if a Disconnect+Connect cycle would help when the radio sits
+      // in IsConnected=True but IsOperational=False for too long.  Default
+      // False -- safe for radios where the two dimensions are independent
+      // (e.g. Flex: TCP up vs SmartSDR slice valid).  TIcomRadio overrides
+      // to True because Icom's "operational" gap means the multi-step
+      // handshake stalled and a fresh AYH is the only recovery.
+      function GetCanRecycleOnStuckHandshake: boolean; virtual;
       function GetAuthFailed: boolean; virtual;
       function BandToFreq(band: TRadioBand): LongInt;  // Map band enum to typical calling frequency
 
@@ -195,6 +202,7 @@ Type TNetRadioBase = class(TObject)
       property IsReceiving: boolean read GetIsReceiving;
       property IsConnected: boolean read GetIsConnected;
       property IsOperational: boolean read GetIsOperational;
+      property CanRecycleOnStuckHandshake: boolean read GetCanRecycleOnStuckHandshake;
       property AuthFailed: boolean read GetAuthFailed;
       property IsRITOn[whichVFO: TVFO]: boolean read GetIsRITOn;
       property IsXITOn[whichVFO: TVFO]: boolean read GetIsXITOn;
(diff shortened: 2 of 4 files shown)


n4af:
https://tr4w.net/4.147/tr4w_setup_4_147.10.exe

PR #909, opened 2026-05-15
Issue #908: bandmap not corrupted by inactive-radio status

### LINKED ISSUE #908: Bandmap blanks at startup when an inactive Icom radio reports band before mode
## Symptom
With a primary radio (e.g., K4 on 20m CW) configured as the active radio and an Icom (e.g., IC-7760 on 80m LSB) configured as a second radio:

1. Start TR4W with the bandmap window open and saved spots from the prior session.
2. Spots flash on the bandmap for about 200 ms, then the bandmap goes blank.
3. Spots only reappear after double-clicking a DX cluster spot (which forces a band change).

Symptom disappears if the inactive Icom is powered off at startup, or the second radio is removed from configuration.

## Root cause
Three call sites propagate `FilteredStatus.Band` and `FilteredStatus.Mode` into globals (`ActiveBand`/`ActiveMode`/`BandMapBand`/`BandMapMode`) without checking whether the radio has actually reported real values. `NoBand` and `NoMode` are sentinel values meaning *"radio has not reported yet"* — but the code overwrites valid state with them anyway.

When the inactive Icom's first poll cycle runs at startup, its `FilteredStatus.Band` was already populated (from a partial unsolicited frame, e.g., \$00 push for frequency) but `FilteredStatus.Mode` is still `NoMode`. The Gav-added "follow inactive radio when tuned" code at `uRadioPolling.pas:3138` then writes `BandMapBand := Band80, BandMapMode := NoMode`, which the bandmap filter rejects every spot against.

A dev comment at `uRadioPolling.pas:3091` already flagged this for the active-radio path: *"After reset port, filteredstatus.band is NoBand which is the issue"*.

## Sites
| File | Line | Path |
|---|---|---|
| `uRadioPolling.pas` | 3088 | active radio polling |
| `uRadioPolling.pas` | 3138 | **inactive radio "follow" — confirmed culprit** |
| `LOGWIND.PAS` | 4933 | legacy LogWind inactive-FS path |

## Fix
Sentinel-value guard at all three sites: skip propagation if `FilteredStatus.Band = NoBand` or `FilteredStatus.Mode = NoMode`.

## Open design questions (for separate discussion)

1. **Asymmetric SO2R gating.** `LOGWIND.PAS:4933` is gated on `TwoRadioState <> TwoRadiosDisabled`, but `uRadioPolling.pas:3138` is not gated on anything SO2R-related. Result: with `Two Radio Mode = False`, an Icom polling thread on a second radio still mutates the bandmap. Should `uRadioPolling.pas:3138` also gate on `TwoRadioMode`? (Probably yes.)

2. **Should the bandmap follow the inactive radio at all?** The Gav-added behavior (pre-existing) auto-switches `BandMapBand`/`BandMapMode` to the inactive radio's band/mode whenever the inactive radio is tuned. The right answer likely involves N4AF's input on intended SO2R workflow.

## Repro signature in tr4w.log (with TRACE level)
```
[DisplayBandMap] enter: BandMapBand=3, BandMapMode=0, FCount=212    <- 73 spots shown OK
[SpotsList.Display] filter pass: passed=73
[DisplayBandMap] enter: BandMapBand=1, BandMapMode=4, FCount=212    <- bandmap blanked
[SpotsList.Display] filter pass: passed=0, rejected by Band=164 Mode=9
```

ny4i:
## Summary

1. **Second radio doesn't mess with the bandmap unless SO2R is actually on.** Previously, even with `TWO RADIO MODE = FALSE`, the inactive radio's polling thread could still rewrite `BandMapBand`/`BandMapMode`. Now it can't.

2. **Don't trust a radio's reported band/mode until it has actually reported real values.** `NoBand` and `NoMode` are placeholders meaning "haven't heard from the radio yet" - they used to leak into the bandmap and active-radio globals; now they're filtered out at all three places that propagated them.

Fixes #908.

## Symptom

With a primary radio (e.g., K4 on 20m CW) configured as the active radio and an Icom (e.g., IC-7760 on 80m LSB) configured as a second radio:

1. Start TR4W with bandmap window open and saved spots from the prior session.
2. Spots flash on the bandmap for ~200 ms, then go blank.
3. Spots only reappear after double-clicking a DX cluster spot.

Symptom disappeared if the inactive Icom was powered off at startup.

## Root cause

Three call sites propagated `FilteredStatus.Band`/`Mode` into globals (`ActiveBand`/`ActiveMode`/`BandMapBand`/`BandMapMode`) without checking whether the radio had reported real values yet. The Gav-added "follow inactive radio when tuned" code at `uRadioPolling.pas:3138` additionally was missing the SO2R gate that the older `LOGWIND.PAS:4933` path already had.

When the inactive Icom's first poll cycle ran at startup, its `FilteredStatus.Band` was already populated (Band80, from a partial unsolicited frame) but `FilteredStatus.Mode` was still `NoMode`. The unguarded code wrote `BandMapBand := Band80, BandMapMode := NoMode`, which the bandmap filter rejected every spot against.

## Fix

| File | Change |
|---|---|
| `uRadioPolling.pas:3088` | Sentinel-value guard at active-radio polling - latent same-pattern bug already flagged by a TODO comment ("After reset port, filteredstatus.band is NoBand which is the issue"). |
| `uRadioPolling.pas:3138` | SO2R gate (`TwoRadioMode`) + sentinel-value guard. **The confirmed culprit.** |
| `LOGWIND.PAS:4933` | Sentinel-value guard at the legacy LogWind inactive-FS path. |

Per N4AF (email thread May 15): the "follow inactive radio when tuned" feature is the intended SO2R workflow ("changes bm entries to the new band - assuming it is the s&p radio that changed bands"), so the SO2R gate preserves that case while eliminating the unintended single-radio behavior. N4AF couldn't reproduce the issue because his INI has `BAND MAP ALL BANDS=TRUE`, which masks the symptom (the band filter accepts everything).

## Open follow-up questions (in #908)

1. **Should the bandmap follow the inactive radio at all in SO2R mode?** A future enhancement could add `BAND MAP FOLLOW INACTIVE RADIO` config knob defaulting `TRUE` (preserves Howie's workflow). Tracked in #908 for separate discussion.

2. **Second-bandmap feature.** N4AF said "i dont think it is worth the effort on 2 bm" - declined for now.

## Diagnostic logging added

TRACE-level only (no behavior change at default log levels):

- `[DisplayBandMap]` - entry state of all bandmap globals + SpotsList.Count
- `[SpotsList.Display]` - per-filter rejection counts (NextDup/Band/Mode/DupeFlag/CQ/WARC/MultsOnly/VHF) + listbox population stats
- `[ProcessFilteredStatus]` - polling thread state on every active-radio poll cycle
- `[BMWriter LOGWIND:4935 inactive-FS]` - fires when the legacy LogWind path mutates bandmap from inactive radio

Useful for future bandmap debugging.

## Test plan

- [x] Reproduce original symptom: K4 active on 20m CW, IC-7760 on 80m LSB, bandmap blanks at startup
- [x] Confirm trace logs identify the inactive-polling code path as the culprit
- [x] Apply fix, rebuild
- [x] Confirm bandmap stays populated with K4's 20m CW spots throughout startup (verified in tr4w.log)
- [x] Confirm `[BMWriter]` and inactive-radio `[ProcessFilteredStatus]` no longer trigger bandmap mutation
- [ ] Verify the "follow inactive radio when tuned" SO2R workflow still works for users with `TWO RADIO MODE = TRUE` (deferred to N4AF testing on K3 SO2R setup)

### DIFF
--- a/tr4w/src/trdos/LOGWIND.PAS
+++ b/tr4w/src/trdos/LOGWIND.PAS
@@ -4466,8 +4466,41 @@ end;
 
 procedure DisplayBandMap;
 begin
-  if not BandMapEnable then Exit;
-  if not tWindowsExist(tw_BANDMAPWINDOW_INDEX) then Exit;
+  // Diagnostic: trace every entry into DisplayBandMap so we can see who
+  // is calling and what state the gates are in.  Useful for tracking
+  // down "bandmap clears unexpectedly" symptoms (issue: bandmap-startup-clear).
+  // Booleans rendered as 0/1 via Ord() to avoid pulling SysUtils.BoolToStr
+  // through this unit's tangled uses chain.
+  logger.Trace('[DisplayBandMap] enter: ' +
+               'BandMapEnable=%d, BandMapWindowExists=%d, ' +
+               'BandMapPreventRefresh=%d, ' +
+               'BandMapBand=%d, BandMapMode=%d, ' +
+               'BandMapAllBands=%d, BandMapAllModes=%d, ' +
+               'BandMapDupeDisplay=%d, BandMapDisplayCQ=%d, ' +
+               'BandMapMultsOnly=%d, WARCBandsEnabled=%d, VHFBandsEnabled=%d, ' +
+               'FCount=%d',
+    [Ord(BandMapEnable),
+     Ord(tWindowsExist(tw_BANDMAPWINDOW_INDEX)),
+     Ord(BandMapPreventRefresh),
+     Ord(BandMapBand), Ord(BandMapMode),
+     Ord(BandMapAllBands),
+     Ord(BandMapAllModes),
+     Ord(BandMapDupeDisplay),
+     Ord(BandMapDisplayCQ),
+     Ord(BandMapMultsOnly),
+     Ord(WARCBandsEnabled),
+     Ord(VHFBandsEnabled),
+     SpotsList.Count]);
+  if not BandMapEnable then
+    begin
+    logger.Trace('[DisplayBandMap] exit: BandMapEnable=False');
+    Exit;
+    end;
+  if not tWindowsExist(tw_BANDMAPWINDOW_INDEX) then
+    begin
+    logger.Trace('[DisplayBandMap] exit: bandmap window does not exist');
+    Exit;
+    end;
 //  if  BandMapPreventRefresh then  Exit;             // GAV  4.45.6
   SpotsList.Display;
   end;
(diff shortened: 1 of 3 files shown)


n4af:
https://tr4w.net/4.147/tr4w_setup_4_147.11.exe

PR #916, opened 2026-05-17
Auto-sync server log + 3830 Score report + Edit Cabrillo Summary + label widening + ADIF CONTEST_ID fix

### LINKED ISSUE #912: Optional auto-synchronize on log mismatch when reconnecting to TR4WServer
## Context

When a TR4W client (re)connects to TR4WServer and the local log's CRC32 differs from the server's, the "Difference in logs" dialog appears and waits for the operator to choose **Synchronize** / **Exit** / **Clear all logs in network**.

In real multi-op operation this dialog fires every time a client reconnects after even a brief disconnect (network blip, laptop sleep, manual server restart) — because the server kept logging new QSOs while the client was offline, so the CRC necessarily differs. Operators report **always** clicking Synchronize, every time, which makes the dialog more obstacle than safeguard.

Pairs with the same problem class addressed in #910 / PR #911 (TR4WSERVER disconnect modal → toast).

## Why this is a TR4W client-side change

- `TLogFileInformation` is a fixed 19-byte packed record with no spare field to carry an "auto-sync" hint from the server.
- Even if a hint existed, the action (overwriting the local log with the server's) happens on the client; the server cannot perform it remotely.
- No protocol change needed.

## What exists today

- **Trigger**: `uNet.pas:838-862` `ProcessServerLogInfo`. Comparison is:
  ```pascal
  if s^.liLocalCRC32 <> s^.liSeverCRC32 then IdenticalLogs := False;
  ```
  Contest is displayed in the dialog but is **not** part of the trigger.
- **Dialog**: `CreateModalDialog(220, 110, tr4whandle, @LogCompareDlgProc, integer(s))` on mismatch. The dialog's **Synchronize** button pulls the full `SERVERLOG.TRW` from the server's sync port (`PORT + 1`).

## Proposed change

Add a new TR4W client-side boolean config flag, default `FALSE` = current behavior. Suggested name (prefixed with `SERVER` to group with the existing `SERVER ADDRESS`, `SERVER PORT`, `SERVER PASSWORD` parameters):

```
SERVER AUTO SYNCHRONIZE LOG ON CONNECT = TRUE
```

Default-off keeps existing behavior; opt-in for operators who already always click Synchronize.

In `ProcessServerLogInfo`, when `IdenticalLogs = False` and the new flag is true, invoke the same code path the **Synchronize** button triggers and skip `CreateModalDialog` entirely.

Suggested log line on auto-sync (writes to `tr4w.log`):

```pascal
logger.Info('Auto-synchronizing local log from server (CRC mismatch: local %x, server %x)', [s^.liLocalCRC32, s^.liSeverCRC32]);
```

Plus a transient `QuickDisplay` toast confirming the sync happened, so the operator sees the action without being interrupted.

## Out of scope

- **Contest mismatch**: not currently in the trigger, intentionally left alone. If contests differ, the existing dialog already shows it; auto-sync would silently overwrite which could surprise an operator who connected to the wrong server. If we ever want to be stricter, the natural check is:
  ```pascal
  if (s^.liLocalCRC32 <> s^.liSeverCRC32) and (LocalContest = s^.liContest)
     then auto-sync
     else still show the dialog
  ```
  Probably not worth the complexity until someone hits the wrong-server case.
- **Direction**: this only handles "server is authoritative, pull from server." TR4W's existing Synchronize button is one-directional in the same way; any "merge local into server" feature would be a separate, much larger change.

## Server compatibility

No server change needed for this feature — auto-sync logic is purely client-side. Both the original Delphi TR4WServer and the Python reimplementation at `c:\projects\TR4WServer` work as-is.

If the protocol ever does grow an auto-sync hint field, both servers would need a coordinated update — but that's not in scope here.

## Test plan

- [ ] With flag = FALSE (default): "Difference in logs" dialog still appears on CRC mismatch — verifies no regression.
- [ ] With flag = TRUE: dialog suppressed; `tr4w.log` shows the `Auto-synchronizing local log from server` line; local log is overwritten with server's copy.
- [ ] With flag = TRUE and identical logs: no dialog, no auto-sync log line (existing `TC_SERVERANDLOCALLOGSAREIDENTICAL` toast still shows).
- [ ] Reconnect after a brief disconnect during a live contest with server still logging QSOs: client log auto-syncs without operator action.

### LINKED ISSUE #914: Edit access to Cabrillo Summary / REPORT fields after contest setup
## Context

The Cabrillo Summary fields (\`_OPERATORS\`, \`_CATEGORY-POWER\`, \`_CATEGORY-OPERATOR\`, \`_CATEGORY-MODE\`, \`_CATEGORY-STATION\`, \`_LOCATION\`, \`_CLUB\`, etc.) live under the \`[REPORT]\` section of \`tr4w.ini\` and are read/written by the Cabrillo Summary dialog in \`uCbrSum.pas\`.

Operators get one chance to set these when first creating a contest. After that, the only way to edit them is to start the Cabrillo file generation (File -> Export -> Cabrillo or Ctrl+Alt+B), which re-opens the dialog. There's no menu option to edit these fields without committing to generating a Cabrillo file.

## Why this matters now

Surfaced by Issue (TBD - 3830 Scores report) which reads from the same \`[REPORT]\` section to populate Call Used / Operators / Class / Power in the 3830 quick-summary report. If those fields are stale or empty, the 3830 report has blanks the operator must fill in manually each time.

The deeper UX gap exists independent of 3830: an operator who realises mid-contest that they set CATEGORY-POWER wrong (say, LOW when they meant HIGH) has to either edit \`tr4w.ini\` by hand, or go through the Cabrillo file-generation flow just to fix one field.

## Proposed fix

New menu item, e.g. \`Tools -> Edit Cabrillo Summary\` or \`File -> Edit Station Info\`. Opens the existing \`OpenStationInformationWindow\` dialog but with a callback that just writes the INI values and returns (no Cabrillo file generated, no \`CreateCabrilloFile\` invoked).

\`OpenStationInformationWindow\` already exists and takes a callback pointer; just need to provide a no-op callback that saves the fields and returns, plus the menu wiring.

## Related

- Issue (TBD - dialog label clipping): label column in the same dialog is too narrow; \`CATEGORY-OPERATOR\` and \`CATEGORY-TRANSMITTER\` labels are visibly clipped. Should ideally be fixed together since both touch the same dialog code.

### LINKED ISSUE #915: New Contest / Open Config dialog: CATEGORY-* labels clipped on the left
## Symptom

In the \"Open configuration file or start a new contest\" dialog (and likely the related Cabrillo Summary dialog, which shares layout code in \`uCbrSum.pas\`), several right-aligned CATEGORY-* labels are clipped off the left because the label column is too narrow:

- \`CATEGORY-OPERATOR\` shows as \"ATEGORY-OPERATOR\" (leading \`C\` clipped)
- \`CATEGORY-TRANSMITTER\` shows as \"EGORY-TRANSMITTER\" (leading \`CAT\` clipped)

Other labels in the column (\`CATEGORY-OVERLAY\`, \`CATEGORY-ASSISTED\`, \`CATEGORY-BAND\`, \`CATEGORY-MODE\`, \`CATEGORY-POWER\`) fit because they are shorter strings.

## Root cause

Right-aligned label control width hard-coded narrower than the longest string. Likely in \`uCbrSum.pas\` (\`CreateCabrilloDlgProc\`) where each \`tCreateEditWindow\` / label control is positioned with fixed pixel offsets.

## Proposed fix

Widen the label column to fit \`CATEGORY-TRANSMITTER\` (the longest string at ~20 chars). Either:

1. Increase the static label width and shift everything (edit/dropdown controls) right by the same amount.
2. Switch the label-control style to \`SS_LEFT\` instead of \`SS_RIGHT\` so clipped text overflows to the right (visually less clean but no layout shift).

Option 1 is the correct fix.

## Related

- Issue (TBD - Edit access to Cabrillo Summary / REPORT fields): both touch the same dialog code in \`uCbrSum.pas\`, may be sensible to fix in one PR.

ny4i:
> ## NOTE TO HOWIE - ONE-TIME ACTION AFTER PULL
>
> After you pull this merge, run this once on your clone to silence cty.dat from \`git status\`:
>
> \`\`\`
> git -C /c/TR4W update-index --skip-worktree tr4w/target/cty.dat
> \`\`\`
>
> Why: cty.dat is rewritten by TR4W at runtime so it always shows as modified.  This PR's gitignore change blocks \`git add\` of cty.dat (so nobody accidentally commits a runtime-modified copy), but already-tracked files still appear in \`git status\` as modified.  The \`--skip-worktree\` flag is per-clone (each developer sets it once on their own machine) and tells your local git to ignore changes to that specific file.  If you ever genuinely need to update the master copy, \`git add -f tr4w/target/cty.dat\` is the escape.
>
> The CHANGES + RELEASE_NOTES commit that added the gitignore update (\`932d89d\` on master) explains this too.

---

## Summary

Four commits across three GitHub issues plus a drive-by ADIF fix.  Two commits chosen for rollback granularity per discussion - shared menu/file infrastructure between 3830 Score and #914 keeps them together; #915 is purely independent and gets its own commit.

### What's in each commit

1. **\`d3e08c5\` Restore CONTEST_ID fallback to ContestTypeSA in ADIF export**
   When \`ContestsArray[Contest].ADIFName\` is empty (true for 156 contests including all CQ-WW / CQ-WPX / ARRL-DX), the ADIF export was emitting no CONTEST_ID at all.  Now falls back to \`ContestTypeSA[Contest]\` like the three other code sites that already do this (\`LOGSUBS2.PAS:2782\`, \`uGetScores.pas:435\`, \`uGetScores.pas:564\`).  Single-line fix covers all 156 contests.  This fallback existed before Issue #887's extraction of the ADIF export from PostUnit/MainUnit into uADIF.pas - it didn't survive the move.

2. **\`3cf72ee\` Fixes #912 - SERVER AUTO SYNCHRONIZE LOG ON CONNECT**
   New boolean config flag (default FALSE).  When TRUE, reconnecting to TR4WServer with a CRC-mismatched log auto-syncs from the server's copy in the background instead of popping the \"Difference in logs\" modal.  Headless implementation: worker thread does the download, marshals to UI thread via SendMessage (\`WM_USER_HEADLESS_SYNC_REPLACE\`) for \`LoadinLog\`'s ListView access.  Safety guard: contest mismatch still falls through to the existing dialog so the wrong-server case still warns.

3. **\`d26d664\` Add 3830 Score report + Fixes #914 Edit Cabrillo Summary access**

   **3830 Score report**: new menu item File -> Reports -> 3830 Score generates \`<logname>_3830Score.txt\` matching the 3830scores.com submission form layout.  Output adapts to contest properties:
   - QSO table: per-band rows (always 160/80/40/20/15/10/6/2 + VHF rows if non-zero); mode columns appear only for modes with QSOs
   - Per-band Mults column iff \`MultByBand = TRUE\` (CQ-WW style)
   - Bottom Mults summary: per-mode lines + Total if \`MultByMode = TRUE\` (FQP/NAQP style); single \`Mults: N\` otherwise
   - Metadata header (Call Used / Operators / Class / Power) read silently from \`tr4w.ini [REPORT]\`; no dialog popup

   **Fixes #914 Edit Cabrillo Summary**: new Tools -> Edit Cabrillo Summary... menu opens the existing dialog standalone (passes \`lParam=0\` so the callback is nil; OK then takes the existing save-and-close path).  Operator can edit \`[REPORT]\` fields post-contest-setup without going through Cabrillo file generation.  Surfaced by the 3830 report's need for those values.

4. **\`c696477\` Fixes #915 - Widen CATEGORY-* labels in New Contest dialog**
   \`CATEGORY-OPERATOR\` and \`CATEGORY-TRANSMITTER\` were clipped on the left because the right-aligned label column was too narrow.  Shrunk the left listbox column by 30 px (280 -> 250), widened the right-side labels by the same 30 px (x=300 -> 270, width 148 -> 178).  Right edges of labels unchanged; dropdown positions unchanged.

## Files

| File | Touched by |
|---|---|
| \`src/uADIF.pas\` | ADIF CONTEST_ID fallback |
| \`src/uNet.pas\`, \`src/uCFG.pas\`, \`src/uGetServerLog.pas\`, \`tr4w.dpr\`, 11 lang files, \`docs/NETWORK_LOG_AUTO_SYNC.md\` | #912 |
| \`src/trdos/PostUnit.PAS\` | 3830 Score report |
| \`src/uCbrSum.pas\` | #914 nil-callback handling |
| \`src/uMenu.pas\`, \`src/VC.pas\`, \`src/MainUnit.pas\` | 3830 + #914 menu wiring |
| \`src/uNewContest.pas\` | #915 |

## Test plan

- [x] ADIF CONTEST_ID - exported CQ-WPX-SSB log, verified \`<CONTEST_ID:10>CQ-WPX-SSB\` present in each record
- [x] #912 SERVER AUTO SYNCHRONIZE - tested with TR4WServer stop/restart; sync happens headlessly, no modal
- [x] 3830 Score report - tested with CQ-WPX-SSB log; metadata block, band table, single-Mults total all correct
- [x] #914 Edit Cabrillo Summary - opens dialog, edits persist to \`tr4w.ini\`, no Cabrillo file generated
- [x] #915 label widening - visual confirmation in New Contest dialog
- [ ] @n4af smoke test on K3 SO2R setup before merging

## Issues this PR closes on merge

- Fixes #912
- Fixes #914
- Fixes #915

### DIFF
--- a/docs/NETWORK_LOG_AUTO_SYNC.md
+++ b/docs/NETWORK_LOG_AUTO_SYNC.md
@@ -1,0 +1,82 @@
+# Feature request: optional auto-synchronize on log mismatch
+
+## Context
+
+When a TR4W client (re)connects to TR4WServer and the local log's CRC32 differs
+from the server's, the "Difference in logs" dialog appears and waits for the
+operator to choose Synchronize / Exit / Clear all logs in network.
+
+In real multi-op operation this dialog fires every time a client reconnects
+after even a brief disconnect (network blip, laptop sleep, manual server
+restart) — because the server kept logging new QSOs while the client was
+offline, so the CRC necessarily differs. Operators report always clicking
+Synchronize, every time, which makes the dialog more obstacle than safeguard.
+
+This is **a TR4W client-side change**. The server cannot drive this decision
+because:
+
+- `TLogFileInformation` is a fixed 19-byte packed record with no spare field
+  to carry an "auto-sync" hint.
+- Even if a hint existed, the action (overwriting the local log with the
+  server's) happens on the client; the server cannot perform it remotely.
+
+## What exists today
+
+- Trigger: `uNet.pas:838-862` `ProcessServerLogInfo`. The relevant comparison
+  is `if s^.liLocalCRC32 <> s^.liSeverCRC32 then IdenticalLogs := False;` —
+  contest is displayed in the dialog but is not part of the trigger.
+- Dialog: `CreateModalDialog(220, 110, tr4whandle, @LogCompareDlgProc,
+  integer(s))` on mismatch. The dialog's Synchronize button pulls the full
+  `SERVERLOG.TRW` from the server's sync port (`PORT + 1`).
+
+## Proposed change
+
+Add an INI key under the TR4W network section, e.g.:
+
+```ini
+[NETWORK]
+AUTO SYNCHRONIZE ON CONNECT = 1
+```
+
+In `ProcessServerLogInfo`, when `IdenticalLogs = False` and the new flag is
+true, invoke the same code path the Synchronize button triggers and skip
+`CreateModalDialog` entirely. Default to `0` (current behavior — show the
+dialog) so this is opt-in for the operator who already knows they always
+click Synchronize.
+
+Suggested log line on auto-sync:
+`logger.Info('Auto-synchronizing local log from server (CRC mismatch: local %x, server %x)', [...])`
+so operators can confirm in `tr4w.log` that the sync happened.
+
+## Out of scope
+
+- Contest mismatch: not currently used in the trigger, intentionally left
+  alone. If contests differ, the existing dialog already shows it; auto-sync
+  would silently overwrite which may surprise an operator who connected to
+  the wrong server. If we ever want to be stricter, the natural check is
+  `if (s^.liLocalCRC32 <> s^.liSeverCRC32) and (LocalContest = s^.liContest)
+  then auto-sync else still show the dialog` — but probably not worth the
+  complexity until someone hits the wrong-server case.
+- Direction: this only handles "server is authoritative, pull from server."
+  TR4W's existing Synchronize button is one-directional in the same way;
+  any "merge local into server" feature would be a separate, much larger
+  change.
+
+## Related: Python TR4WServer
+
+The Python reimplementation of TR4WServer (in c:\projects\TR4WServer)
+mirrors the Delphi server's wire protocol exactly. It does not need any
+changes for this feature — the auto-sync logic is purely client-side. If
+the protocol ever does grow an auto-sync hint field, both servers would
+need a coordinated update.
+
+The Python server already implements:
+- Serial-number lockout (matches `tr4wserverUnit.pas` behavior — startup
+  scan of `NumberSent`, per-client `SM_SERIAL_NUMBER_CHANGED` push on
+  connect, increment on reservation).
+- Live per-station status table built from `TStationState` broadcasts
+  (Band / Freq / St / PTT / Qs / Callsign / Op).
+- `LOG RECORD SIZE` is configurable in `tr4wserver.ini` and validated
+  against the on-disk `SERVERLOG.TRW`, so a future TR4W release that
+  changes `SizeOf(ContestExchange)` produces a clear startup error rather
+  than silent corruption.
--- a/tr4w/src/MainUnit.pas
+++ b/tr4w/src/MainUnit.pas
@@ -3600,6 +3600,8 @@ begin
 
     menu_cabrillo: OpenStationInformationWindow(integer(@CreateCabrilloFile));
     menu_summary: OpenStationInformationWindow(integer(@SummarySheet));
+    menu_3830scores: ExportTo3830Scores;  // Issue: 3830 quick-submission report
+    menu_edit_cabrillo_summary: OpenStationInformationWindow(0);  // Issue #914
     menu_export_edi: OpenStationInformationWindow(integer(@ExportToEDI));
 
     menu_scorebyhour: ScoreByHour;
--- a/tr4w/src/VC.pas
+++ b/tr4w/src/VC.pas
@@ -2299,6 +2299,8 @@ const
   menu_import_adif                      = 10018;
   menu_export_edi                       = 10019;
   menu_csv                              = 10020;
+  menu_3830scores                       = 10021;
+  menu_edit_cabrillo_summary            = 10022;  // Issue #914
   menu_options                          = 10100;
   menu_messages                         = 10101;
   menu_other_messages                   = 10102;
--- a/tr4w/src/lang/TR4W_CONSTS_ESP.PAS
+++ b/tr4w/src/lang/TR4W_CONSTS_ESP.PAS
@@ -184,6 +184,7 @@ const
   TC_CONNECTTOTR4WSERVERFAILED          = 'Conexion a TR4WSERVER fallida. Compruebe valor SERVER PASSWORD!!';
   TC_CONNECTEDTO                        = 'Conectando a';
   TC_DISCONNECTEDFROM                   = '** Desconectado de ';
+  TC_AUTOSYNCHRONIZINGLOG               = 'Sincronizando log desde el servidor automaticamente.';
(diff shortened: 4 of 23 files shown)


n4af:
one time cty.dat action completed.
https://tr4w.net/4.147/tr4w_setup_4_147.13.exe

PR #919, opened 2026-05-19
Issues #917, #436: RadioInfo UDP fixes + non-blocking network connect + Tier 1 test groundwork

### LINKED ISSUE #917: IsConnected is missing from the RadioInfo UDP Message
The N1MM RadioInfoXML spec includes a IsConnected field defined as follows:

IsConnected indicates whether a radio is connected. It may take some time for a disconnection (False) to be registered

The XML Field is <IsConnected></IsConnected>.

Implement this to send based on if the radio is connected or not. Handle this in both the network or serial port status.

ny4i:
## Summary

Two correctness fixes for amateur-radio network radio control, plus the start of a unit-test regression net ahead of the Delphi 12 migration.

### Issue #917 — `<IsConnected>` missing from RadioInfo UDP, plus three sibling bugs in the same XML

- **`<IsConnected>` field added** to the N1MM-compatible RadioInfo UDP broadcast (`SendRadioInfoToUDP` in `uRadioPolling.pas`). Source is `rig.RadioDisconnected`, which `pNetworkRadio`'s `SetRadioAlertState` already maintains for every network radio (Icom CI-V/IP, K4 TCP, TS-890 LAN, FlexRadio). Network-side disconnect detection is real-time; serial-attached radios report `True` whenever configured (no liveness signal exists for serial today — separate follow-up).
- **`<RadioNr>` bug**: was keyed on the global `ActiveRadio` instead of the packet's subject `rig`. In SO2R both packets reported the same number; for `RadioNr` to be useful per N1MM spec, the field must identify *which* radio the packet describes. Now `IfThen(rig = @Radio1, 1, 2)`.
- **`<FocusRadioNr>` and `<ActiveRadioNr>` bugs**: hardcoded `'1'`. Now reflect the global `ActiveRadio` so SO2R operator-focus flips show up in the broadcast.

### Issue #436 — TS-890S LAN connect was broken; main thread blocked during all network radio Connect

Two compounding root causes:

1. **Base-class `OnRadioConnected` sent `'ID;'` synchronously inside `socket.Connect`**, before the TS-890 LAN auth opener `##CN;` had a chance to fire from the override. The TS-890 saw an unauthenticated command, dropped the session, and on N2SKH's hardware blanked the bandscope and rebooted. Per N2SKH's pcap, this fires every ~12 s in an endless reconnect cycle.

   Deleted the `'ID;'` send entirely. Each subclass already owns its initial probe — the K4's `Connect` sends a full state-query batch including `ID;`, FlexRadio's `Connect` had a comment specifically documenting that it tolerates this wasted `ID;`, Icom CI-V/HamLib don't trigger this path at all, and the TS-890 path now goes auth-first as the radio requires.

2. **Initial network Connect ran synchronously on the UI thread.** `SetUpRadioInterface` called `tNetObject.Connect` directly, blocking the main thread for up to 10 s (Indy `ConnectTimeout`) whenever the radio was unreachable. Indy's `OnConnected` also fires synchronously inside `socket.Connect`, so the post-connect probe ran on the UI thread too.

   Deferred the Connect to `pNetworkRadio`'s existing reconnect-with-backoff loop. The polling thread already handles `ro.Connect` failures with exponential backoff; now it owns the very first connect as well. No new threading machinery; the disconnected branch in `pNetworkRadio` was already proven for reconnects. `StartupCommand` (configurable per-radio in `.cfg`) is now sent once after the first successful connect via a `StartupCommandSent: Boolean` one-shot on `RadioObject` — race-free with the deferred Connect, and resets on Reset Radio Ports.

### Pre-migration test groundwork (separate concern, deferred to the docs-only commit on master)

While waiting on Ron's TS-890 bench testing, picked four items off the Pre-Migration Interim Roadmap. Same branch for atomicity:

- **`uBandLookup.pas`** (new): `CalculateBandMode(Freq → BandType + ModeType)` extracted from `trdos/tree.pas` (5,292 lines, drags Win32 + TF) into a 47-line unit. `tree.pas` keeps the public symbol as a documented forwarding stub; zero caller changes across the 10 existing call sites. 44 test methods in `uTestBandLookup.pas` covering every band, every mode-split boundary, every band edge, and the 40 m NoMode quirk.
- **`uCabrilloFormat.pas`** (new): `FormatCabrilloFreq` + `FormatCabrilloMode` extracted from `PostUnit.tGenerateLogPortionOfCabrilloFile` (1,226 lines) into a small unit. Arrays `tCabrilloFreqString` + `tCabrilloModeString` moved with them; three call-site files (`PostUnit`, `LOGSUBS2`, `uWSJTX`) updated to `uses uCabrilloFormat`. 21 test methods.
- **`uCRC32.pas` tests** (`uTestCRC32.pas`): 8 tests against published CRC-32/ISO-HDLC reference vectors (`CRC32(\"123456789\") == \$CBF43926` and 7 more). Network packet integrity for TR4WServer; the inline asm bodies are Phase 3 rewrite targets.
- **`utils/utils_math.pas` tests** (`uTestUtilsMath.pas`): 15 tests covering `Tan`, `ArcCos`, `ArcTan2` (two of three use x86 FPTAN/FPATAN asm). Freezes the D7 bit-exact outputs so the Phase 3 Pascal rewrite is provably equivalent. Used by grid-distance scoring (RTC, FQP).
- **DupeAndMultSheet spike** — concluded not instantiable in a test EXE without major extraction (drags LogWind 172k + LogRadio 152k + LogSCP + Tree); finding documented in `docs/tr4w-migration-strategy.md`.
- **Phase 2 + 3 inventories**: 600 inline asm blocks across 74 files (PostUnit alone has 170, 28% of the total). 172 wsprintf calls across 37 files — original migration analysis estimate of 73 was 2.4× too low. Full sorted tables and reproducible grep commands in `docs/PHASE_INVENTORIES.md`.

## Test plan

- [ ] **TS-890 LAN (N2SKH)**: confirm `[TS-890S.HandleAuthMessage] Authenticated successfully` in `tr4w.log`; bandscope no longer blanks; freq/mode appears in TR4W within seconds; no 5–10 s drop cycles. Drop the new `tr4w.exe` into the existing v4.147.08 install (no other artifacts changed).
- [x] **IC-7760 / K4 bench (N4AF)**: regression — connect with radio on works as before; connect with radio off no longer freezes the UI for ~10 s; turning radio on reconnects within seconds; magenta indicator behaves as expected; `StartupCommand` (if configured) fires once after first connect, again after Reset Radio Ports.
- [x] **N1MM RadioInfo consumer**: `<IsConnected>True</IsConnected>` while radio is up, `False` within ~3 s of powering off (network Icoms). In SO2R, each radio's packet carries its own `RadioNr`; `FocusRadioNr`/`ActiveRadioNr` flip when active radio changes.
- [x] **Cabrillo regression**: any contest log export. Output should be byte-identical to v4.147.13 — the freq/mode formatter extraction preserves behavior verbatim (no inline call-site replacement; the array values + branch logic stayed the same).
- [ ] **FlexRadio**: connect, log a QSO, exit. Regression check that the absent base-class `ID;` doesn't change anything visible (was previously tolerated as a known wasted error round-trip).
- [x] **TR4W rebuilds clean**, unit tests pass: `PASSED: 773  FAILED: 0`.

## Issues

Closes #917 (RadioInfo `<IsConnected>` field — plus three sibling bugs surfaced in the same XML block that aren't separate issues but are documented in the commit messages and CHANGES on master).

Partially addresses #436 (TS-890S LAN protocol bug + main-thread network connect). Per N4AF's note, the 890 network protocol may need further refinement as additional issues are discovered; those will be follow-up PRs.

## Commits

- `2d64e6d` Issues #917, #436: RadioInfo UDP fixes + non-blocking network radio connect
- `b192bc8` docs/tr4w-migration-strategy: Tier 1 status, extraction pattern, interim roadmap *(already on master via cherry-pick `1a70522`)*
- `38e853a` Tier 1 test coverage + Phase 2/3 inventories + extraction pattern in practice
- `a42dff8` Tier 1 tests: uCRC32 reference vectors + utils_math FPU asm freeze

### DIFF
--- a/docs/PHASE_INVENTORIES.md
+++ b/docs/PHASE_INVENTORIES.md
@@ -1,0 +1,228 @@
+# Phase 2 & 3 Migration Inventories
+
+Pre-migration inventory artifacts that feed the Phase 2 (Unicode) and
+Phase 3 (64-bit) checklists. See `docs/tr4w-migration-strategy.md`
+items 10 and 11 in the Pre-Migration Interim Roadmap.
+
+Generated 2026-05-19. Counts come from a static `grep` over `tr4w/src/`,
+excluding `~`, `.bakup`, and `.bad` working files.
+
+---
+
+## Phase 3 — Inline `asm` Block Inventory
+
+**Total: 600 `asm` blocks across 74 files.**
+
+Inline x86 assembly is inaccessible in 64-bit Delphi compilation. Every
+block must be either:
+
+- Replaced with equivalent Pascal code, or
+- Gated with `{$IFDEF CPUX86}…{$ENDIF}` plus a 64-bit Pascal alternative.
+
+The top of the list is where to start — `PostUnit.PAS` alone is 28% of
+all blocks in the project.
+
+### Files sorted by count (descending)
+
+| File | `asm` blocks |
+|------|--------------|
+| tr4w/src/trdos/PostUnit.PAS | 170 |
+| tr4w/src/utils/SysUtils.pas | 58 |
+| tr4w/src/MainUnit.pas | 56 |
+| tr4w/src/MySU.pas | 39 |
+| tr4w/src/uVariants.pas | 24 |
+| tr4w/src/TF.pas | 20 |
+| tr4w/src/uLogCompare.pas | 17 |
+| tr4w/src/uQTCS.pas | 14 |
+| tr4w/src/trdos/LOGDVP.PAS | 10 |
+| tr4w/src/uWinKey.pas | 8 |
+| tr4w/src/trdos/LOGSTUFF.PAS | 8 |
+| tr4w/src/uQTCR.pas | 7 |
+| tr4w/src/uGetScores.pas | 6 |
+| tr4w/src/uComObj.pas | 6 |
+| tr4w/src/trdos/tree.pas | 6 |
+| tr4w/src/trdos/_JCtrl1.pas | 6 |
+| tr4w/src/trdos/LOGWAE.PAS | 6 |
+| tr4w/src/trdos/JCtrl1.pas | 6 |
+| tr4w/src/uTelnet.pas | 5 |
+| tr4w/src/uAltP.pas | 5 |
+| tr4w/src/trdos/LogCfg.pas | 5 |
+| tr4w/src/tr4wserverUnit.pas | 5 |
+| tr4w/src/uRemMults.pas | 4 |
+| tr4w/src/uRadioPolling.pas | 4 |
+| tr4w/src/uProcessCommand.pas | 4 |
+| tr4w/src/uNewContest.pas | 4 |
+| tr4w/src/uIntercom.pas | 4 |
+| tr4w/src/uEditMessage.pas | 4 |
+| tr4w/src/uCFG.pas | 4 |
+| tr4w/src/trdos/LOGWIND.PAS | 4 |
+| tr4w/src/trdos/LOGSUBS2.PAS | 4 |
+| tr4w/src/jwamswsock.pas | 4 |
+| tr4w/src/exportto_trlog.pas | 4 |
+| tr4w/src/uRemMults_Zone.pas | 3 |
+| tr4w/src/uCallsigns.pas | 3 |
+| tr4w/src/uCRC32.pas | 3 |
+| tr4w/src/trdos/LOGK1EA.PAS | 3 |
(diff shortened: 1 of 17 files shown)


n4af:
https://tr4w.net/4.147/tr4w_setup_4_147.14.exe

PR #922, opened 2026-05-24
Uninit-locals fixes (CW-by-CAT + bandmap), TS-890 LAN keepalive + handshake, build zip step

### LINKED ISSUE #1: Testing issues
Test

### LINKED ISSUE #2: Connecting to dxc.ww1r.com:7300 causes crash
When I select dxc.ww1r.com:7300 from the drop down in the DX Cluster window, the program crashes right after the connect. I suspect the program is receiving something from the cluster it does not expect. I will check this out with the debugger and see what we are receiving.

ny4i:
## Summary

Four independent improvements gathered onto one branch:

### 1. Fix uninitialized-locals bug in `RadioObject.SendCW` — Kenwood/Elecraft CW-by-CAT
`charProcessed` and `sendNow` were declared but never initialized at function entry. Delphi 7 does not zero stack-allocated locals, so they carried whatever was on the stack from prior calls. When `charProcessed` happened to be non-zero on entry, the `if not charProcessed` guard at the bottom of the rtKenwood case skipped the append, so individual letters never made it into `CWByCATBuffer`. By the time the terminator arrived the buffer was empty and no `KY` command went out. Symptom: per-char `[AddStringToBuffer] Adding (N) to CW Buffer` debug logs appear, then `Assembling CW message to send`, then nothing. Affects all Kenwood and Elecraft radios using CW-by-CAT. Surfaced testing RTC mode with a K4 over TCP.

### 2. Fix uninitialized `CurrentCursorPos` in bandmap repaint (`LOGWIND.PAS`)
Same signature as #1. Found via a code audit after the SendCW fix. `CurrentCursorPos` was only assigned inside the band/mode/entry loops when a row matched the previously-selected cursor entry, then passed unconditionally to `tLB_SETCURSEL` at the bottom of the procedure. When no row matched (entry expired, dupe filter toggled, frequency walked off, bandmap repopulated) the listbox cursor was set to stack garbage. Fix: initialise to `-1` (LB_SETCURSEL "no selection") right after the cursor-pointer read.

### 3. TS-890 LAN: `PS;` keepalive every 5s + wait for `##TI;` before init + parse TB/FT/RT/XT
Three fixes for the Kenwood TS-890 LAN protocol per the [open890 LAN-HOWTO](https://github.com/open890/kenwood-ts890/blob/main/LAN-HOWTO.md) and an audit against `simts890.c`:
- **`PS;` keepalive every 5 seconds.** The TS-890 closes its LAN TCP connection after 10 seconds of client silence. AI2 only pushes state outbound; the radio needs traffic *from* us. Kenwood's own ARCP sends `PS;` every 5 s. Match that with `requiresPolling=True`, `pollingInterval=5000`, `PollRadioState` sending `PS;`, and a `PS` dispatch case that no-ops the reply (the round-trip itself is the keepalive). Without this, the radio dropped us ~10 s after auth.
- **Wait for `##TI;` before sending CAT.** Per the HOWTO, after `##ID1;` the radio batches `##ID1;##UE1;##TI1;` and only considers itself CAT-ready after `##TI1;`. We were firing `InitializeAfterAuth` on `##ID1;`. New `ksWaitingForTI` auth state handles `##UE` (log + ignore) and `##TI` (promote to authenticated, run init).
- **Parse TB/FT/RT/XT responses.** These four queries were sent in the init burst but their responses landed in the `Unhandled reply` trace bucket. Now parsed: `TB` → `localSplitEnabled`, `RT` → `RITState`, `XT` → `XITState`. `FT` (TX VFO) logged for visibility until a per-radio TX-VFO field exists.

### 4. `FullBuild.ps1`: zip the exe into `target/dist/` on successful build
Drops `tr4w-<version>-<branch>-<YYYYMMDD-HHMMSS>.zip` next to `target/tr4w.exe`. Use case: iterative bug-hunting with off-site testers (radio bugs where the developer doesn't have the hardware locally) where three rounds of `tr4w.exe` attachments in an inbox get ambiguous. `target/dist/` is already covered by the existing `tr4w/target/*` gitignore.

## Commits

- `5b75b94` Fix uninitialized-locals bug in `RadioObject.SendCW`
- `51bfa67` Fix uninitialized `CurrentCursorPos` in bandmap repaint
- `21df4a1` TS-890 LAN: keepalive + `##TI` wait + TB/FT/RT/XT parsers
- `8ef238a` `FullBuild.ps1`: post-build zip step

## Test plan

- [x] `FullBuild.ps1` runs all 773 unit tests + main build green
- [x] Local CW-by-CAT verified working on K4/TCP — `KY <call>;` reaches the radio (was completely broken before fix #1)
- [ ] TS-890 LAN fixes (#3) tested with real hardware by N2SKH — the radio he was testing previously dropped ~20s after auth; with `PS;` keepalive the connection should stay up and state queries should populate VFO B / mode / split / RIT / XIT
- [ ] Bandmap repaint cursor (fix #2) tested on a live log — selection no longer jumps after a repaint when previous selection has aged out / filter changed
- [x] `target/dist/<versioned-zip>.zip` produced on each successful build, ignored by git

### DIFF
--- a/tr4w/FullBuild.ps1
+++ b/tr4w/FullBuild.ps1
@@ -75,6 +75,50 @@ if ($result -eq 0) {
         Write-Host "  Size: $($exeInfo.Length) bytes" -ForegroundColor White
         Write-Host "  Modified: $($exeInfo.LastWriteTime)" -ForegroundColor White
         Write-Host ""
+
+        # ---------------------------------------------------------------
+        # Step 3: Zip the exe for distribution.
+        # The version + branch + timestamp filename means whack-a-mole test
+        # cycles with off-site testers never end up with two ambiguous
+        # "tr4w.exe" attachments sitting in the inbox.
+        # ---------------------------------------------------------------
+        $DIST_DIR = "C:\TR4W\tr4w\target\dist"
+        New-Item -ItemType Directory -Force -Path $DIST_DIR | Out-Null
+
+        # Extract version from Version.pas (e.g. 4.147.15)
+        $versionLine = Select-String -Path "C:\TR4W\tr4w\src\Version.pas" `
+                                     -Pattern "TR4W_CURRENTVERSION_NUMBER\s*=\s*'([^']+)'" `
+                                     | Select-Object -First 1
+        if ($versionLine -and $versionLine.Matches[0].Groups[1].Value) {
+            $version = $versionLine.Matches[0].Groups[1].Value
+        } else {
+            $version = "unknown"
+        }
+
+        # Current git branch (slash-safe for filenames)
+        $branchRaw = (git -C "C:\TR4W" rev-parse --abbrev-ref HEAD 2>$null)
+        if ($LASTEXITCODE -ne 0 -or -not $branchRaw) {
+            $branch = "nobranch"
+        } else {
+            $branch = $branchRaw.Trim() -replace '[/\\:*?"<>|]', '-'
+        }
+
+        $stamp   = Get-Date -Format "yyyyMMdd-HHmmss"
+        $zipName = "tr4w-$version-$branch-$stamp.zip"
+        $zipPath = Join-Path $DIST_DIR $zipName
+
+        Write-Host "=== Packaging EXE ===" -ForegroundColor Cyan
+        Write-Host "Archive: $zipPath" -ForegroundColor Yellow
+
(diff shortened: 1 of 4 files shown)


n4af:
https://tr4w.net/4.147/tr4w_setup_4.147.17.exe

PR #988, opened 2026-06-07
Issue #986: open previewed files in the default text editor, not Notepad

### LINKED ISSUE #986: Open to default document for txt editors when opening ADI amd CAB file preview
When previewing the exported files such as ADI and Cabrillo, instead of forcing it to Notepad, use the default program for a text editor on the system.

Query Windows for the executable associated with `.txt`, then launch that executable with your `.adi` file as a parameter. `AssocQueryString` can return the executable registered for a given extension, including `.txt`. [[stackoverflow](https://stackoverflow.com/questions/2577213/how-can-i-get-the-name-of-the-program-associated-with-a-file-extension-using-del)](https://stackoverflow.com/questions/2577213/how-can-i-get-the-name-of-the-program-associated-with-a-file-extension-using-del)

## Delphi example

This version gets the default editor for `.txt` and uses it to open your `.adi` file. [[physiologyweb](https://www.physiologyweb.com/delphi_code_examples/open_application_and_pass_parameters.html)](https://www.physiologyweb.com/delphi_code_examples/open_application_and_pass_parameters.html)

```pascal
uses
  Winapi.Windows,
  Winapi.ShellAPI,
  Winapi.ShlObj,
  System.SysUtils;

function GetAssociatedExecutable(const Ext: string): string;
var
  Len: DWORD;
begin
  Result := '';
  Len := 0;

  // First call gets required buffer size
  AssocQueryString(
    ASSOCF_NONE,
    ASSOCSTR_EXECUTABLE,
    PChar(Ext),
    nil,
    nil,
    @Len
  );

  if Len > 0 then
  begin
    SetLength(Result, Len - 1);
    if AssocQueryString(
         ASSOCF_NONE,
         ASSOCSTR_EXECUTABLE,
         PChar(Ext),
         nil,
         PChar(Result),
         @Len
       ) <> S_OK then
      Result := '';
  end;
end;

procedure OpenADIWithTextEditor(const FileName: string);
var
  EditorExe: string;
  Rc: HINST;
begin
  EditorExe := GetAssociatedExecutable('.txt');
  if EditorExe = '' then
    raise Exception.Create('No editor is associated with .txt files.');

  Rc := ShellExecute(
          Application.Handle,
          nil,
          PChar(EditorExe),
          PChar('"' + FileName + '"'),
          nil,
          SW_SHOWNORMAL);

  if NativeInt(Rc) <= 32 then
    raise Exception.CreateFmt('Unable to open file. ShellExecute error %d', [NativeInt(Rc)]);
end;
```

## Why this works

`AssocQueryString(..., ASSOCSTR_EXECUTABLE, '.txt', ...)` asks Windows which executable is registered to handle `.txt` files. Then `ShellExecute` launches that executable directly and passes your `.adi` filename as a quoted command-line parameter. [[learn.microsoft](https://learn.microsoft.com/en-us/windows/win32/shell/launch)](https://learn.microsoft.com/en-us/windows/win32/shell/launch)

Quoting the filename is important because the `.adi` path may contain spaces. When you launch the editor executable directly, the file path is now just a command-line argument, so it needs proper quotes. [[community.notepad-plus-plus](https://community.notepad-plus-plus.org/topic/17151/invoking-notepad-from-another-application-using-shellexecex-issue-with-7-6-3)](https://community.notepad-plus-plus.org/topic/17151/invoking-notepad-from-another-application-using-shellexecex-issue-with-7-6-3)

## Notes

This gives you the user’s preferred text editor for `.txt`, but it does assume that editor accepts a filename on its command line, which most normal text editors do. In practice that works for Notepad, Notepad++, VS Code, and most editors a user would reasonably assign to `.txt`. [[physiologyweb](https://www.physiologyweb.com/delphi_code_examples/open_application_and_pass_parameters.html)](https://www.physiologyweb.com/delphi_code_examples/open_application_and_pass_parameters.html)

If you want the most shell-correct version, you can also query `ASSOCSTR_COMMAND` for `.txt` instead of just the executable, because some associations may define a more complex command template than simple `editor.exe "%1"`. `ASSOCSTR_EXECUTABLE` is usually enough, but `ASSOCSTR_COMMAND` is the more exact association data. [[stackoverflow](https://stackoverflow.com/questions/2577213/how-can-i-get-the-name-of-the-program-associated-with-a-file-extension-using-del)](https://stackoverflow.com/questions/2577213/how-can-i-get-the-name-of-the-program-associated-with-a-file-extension-using-del)

## Simpler fallback

If you do not care specifically about the `.txt` editor and just want “open as text somehow,” another practical option is to create a temporary `.txt` copy of the ADI content and `ShellExecute` that file directly. That lets Windows use the full `.txt` association path with no assumptions. [[blogs.embarcadero](https://blogs.embarcadero.com/opening-a-file-with-the-default-program-on-windows-and-mac/)](https://blogs.embarcadero.com/opening-a-file-with-the-default-program-on-windows-and-mac/)

Would you like the `ASSOCSTR_COMMAND` version too, which parses the full registered `.txt` open command?

ny4i:
Closes #986.

## Summary

Previewed export files (and `history.txt`) were always opened in Notepad. This routes every "open in editor" path through the user's **default text editor** — the program registered for the `.txt` extension — instead of hard-coding Notepad.

## Changes

**New shared helper** `OpenInDefaultTextEditor` (`MainUnit.pas`):
- Resolves the executable registered for `.txt` via `AssocQueryStringA(ASSOCF_NONE, ASSOCSTR_EXECUTABLE, '.txt', …)` (declared directly — Delphi 7's ShlwApi import unit doesn't expose it).
- Launches it with `ShellExecute`, passing the file path **quoted** so paths containing spaces work.
- Opens in a **normal** window (`SW_SHOWNORMAL`).
- **Falls back to the original Notepad behavior** if no `.txt` association resolves or the editor fails to launch, so it never regresses on a misconfigured system.
- Uses a local command buffer (not the shared `wsprintfBuffer`) so a caller passing `wsprintfBuffer` can't self-corrupt.

**Call sites routed through the helper:**
- `uFileView.pas` — the file-preview window's File menu (summaries, 3830 report, ADI, Cabrillo — all previewed files).
- `MainUnit.pas` — the `history.txt` menu item.

**Menu label:** `uMenu.pas` — "Open with Notepad" → "Open in text editor".

## Testing

Manually tested on Windows 11 with the `.txt` default set to **Notepad++**:
- Preview files and `history.txt` open in Notepad++ rather than Notepad.
- File paths with spaces open correctly (quoting).
- TR4W passes `SW_SHOWNORMAL`; the only observed maximize came from Notepad++'s own remembered-window-state setting, not the launch flag.

### DIFF
--- a/tr4w/src/MainUnit.pas
+++ b/tr4w/src/MainUnit.pas
@@ -213,6 +213,7 @@ procedure CheckQuestionMark;
 function Get101Window(h: HWND): HWND;
 procedure InvertBooleanCommand(Command: PBoolean);
 procedure RunExplorer(Command: PChar);
+procedure OpenInDefaultTextEditor(FileName: PChar);   // Issue #986
 procedure RunOptionsDialog(f: CFGFunc);
 procedure OpenUrl(url: PChar);
 function ParseADIFRecord(sADIF: string; var exch: ContestExchange): boolean;
@@ -3655,9 +3656,9 @@ begin
 
     menu_historytxt:
       begin
-        // OpenUrl('http://www.tr4w.com/');
-        Format(wsprintfBuffer, 'notepad %shistory.txt', TR4W_PATH_NAME);
-        WinExec(wsprintfBuffer, SW_SHOW);
+        // Issue #986 -- open in the system default text editor, not Notepad.
+        Format(wsprintfBuffer, '%shistory.txt', TR4W_PATH_NAME);
+        OpenInDefaultTextEditor(wsprintfBuffer);
       end;
 
     menu_wiki_rus:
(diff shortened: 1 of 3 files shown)


ny4i:
This closes #988

